PR #3, opened 2022-01-12
WIP: Requirement lists

hikinggrass:


caller:
@tmolitor-stud-tu  you last commit wasn't signed, so we cant merge this... can you fix that?

tmolitor-stud-tu:
It is still WIP (the C++ code generator stuff is still missing), so nothing to merge yet

everestjs/utils.hpp
```diff
@@ -13,52 +13,42 @@ namespace EverestJs {
...
         try {                                                                                                          \
             throw;                                                                                                     \
-        } catch (Napi::Error & e) {                                                                                    \
+        } catch (Napi::Error& e) {                                                                                     \
             try {                                                                                                      \
                 BOOST_THROW_EXCEPTION(boost::enable_error_info(e)                                                      \
                                       << boost::log::BOOST_LOG_VERSION_NAMESPACE::current_scope());                    \
-            } catch (std::exception & e) {                                                                             \
+            } catch (boost::exception& ex) {                                                                           \
```
a-w50, line 20: What is the reason for calling BOOST_THROW_EXCEPTION and catching in again afterwards? Is it only to get the file, line and function information via boost log?

lib/config.cpp
```diff
@@ -733,80 +735,65 @@ void Config::resolve_all_requirements() {
...
 
-            json connection_provides = connection_manifest["provides"][connection_impl_id];
-            json connection_provides_intfs = this->base_interfaces[connection_module_name][connection_impl_id];
-            for (auto& req : requirement.items()) {
-                const auto& requirement_type = req.key();
-                auto& requirement_value = req.value();
+            for (uint64_t connection_num = 0; connection_num < connections.size(); connection_num++) {
+                auto& connection = connections[connection_num];
+                std::string connection_module_id = connection["module_id"];
+                if (!this->main.contains(connection["module_id"])) {
```
a-w50, line 773: why not use the previously defined variable, ie main.contains(connection_module_id) ?

lib/config.cpp
```diff
@@ -733,80 +735,65 @@ void Config::resolve_all_requirements() {
...
+            // check if min_connections and max_connections are fulfilled
+            if (connections.size() < requirement["min_connections"] ||
+                            connections.size() > requirement["max_connections"]) {
                 EVLOG_AND_THROW(EVEXCEPTION(EverestConfigError, "Requirement '", requirement_id, "' of module ",
-                                            printable_identifier(module_id), " not fulfilled: required module ",
-                                            printable_identifier(connection["module_id"]),
-                                            " does not provide a implementation for '", connection_impl_id, "'!"));
+                                            printable_identifier(module_id), " not fulfilled: requirement list does",
+                                            "not have an entry count between ", requirement["min_connections"],
+                                            " and ", requirement["max_connections"], "!"));
```
a-w50, line 767: For the debug message it would be helpful to also see the actual value of connections.size()

lib/config.cpp
```diff
@@ -733,80 +735,65 @@ void Config::resolve_all_requirements() {
...
-                const auto& requirement_type = req.key();
-                auto& requirement_value = req.value();
+            for (uint64_t connection_num = 0; connection_num < connections.size(); connection_num++) {
+                auto& connection = connections[connection_num];
+                std::string connection_module_id = connection["module_id"];
+                if (!this->main.contains(connection["module_id"])) {
+                    EVLOG_AND_THROW(EVEXCEPTION(EverestConfigError, "Requirement '", requirement_id, "' of module ",
+                                                printable_identifier(module_id), " not fulfilled: module id '",
+                                                connection_module_id,
+                                                "' (configured in connection ", connection_num,
```
a-w50, line 777: would rather write "referenced in connection"

lib/config.cpp
```diff
@@ -733,80 +735,65 @@ void Config::resolve_all_requirements() {
...
+                if (provided_intfs.count(requirement_interface) < 1) {
+                    EVLOG_AND_THROW(EVEXCEPTION(
+                        EverestConfigError, "Requirement '", requirement_id, "' of module ",
+                        printable_identifier(module_id), " not fulfilled by connection to module ",
+                        printable_identifier(connection["module_id"], connection_impl_id),
+                        ": required interface '", requirement_interface,
+                        "' is not provided by this implementation!",
+                        " Connected implementation provides interface '",
+                        connection_provides["interface"].get<std::string>(),
+                        "' with parent interfaces: ", json(provided_intfs)));
```
a-w50, line 806: we should definitely switch to libfmt some day

include/framework/everest.hpp
```diff
@@ -76,6 +76,7 @@ class Everest {
     /// requirement_id, the command by the given command name \p cmd_name and the needed arguments by \p args
     ///
     json call_cmd(const std::string& requirement_id, const std::string& cmd_name, json json_args);
+    json call_cmd(const std::string& requirement_id, uint64_t index, const std::string& cmd_name, json json_args);
```
a-w50, line 78: uint64_t? That's what I call "think large"

everestjs/everestjs.cpp
```diff
@@ -107,27 +107,26 @@ static Napi::Value setup_cmd_handler(const std::string& impl_id, const std::stri
...
 
-        auto sub_key = std::make_pair(requirement_id, var_name);
+        auto sub_key = std::make_tuple(requirement_id, index, var_name);
         auto& var_subs = ctx->var_subscriptions;
 
         if (var_subs.find(sub_key) != var_subs.end()) {
             // FIXME (aw): error handling, if var is already subscribed
             EVTHROW(EVEXCEPTION(Everest::EverestApiError, "Subscribing to ", requirement_id, "->", var_name,
-                                " more then once is not yet supported!"));
+                                " more than once is not yet supported!"));
```
a-w50, line 123: good find :)

everestjs/everestjs.cpp
```diff
@@ -375,93 +374,99 @@ static Napi::Value boot_module(const Napi::CallbackInfo& info) {
...
+            auto req_mod_cmds_array = Napi::Array::New(env);
+            for (uint64_t i = 0; i < req_route_list.size(); i++) {
+                auto req_route = req_route_list[i];
+                const std::string& requirement_module_id = req_route["module_id"];
+                const std::string& requirement_impl_id = req_route["implementation_id"];
+                // FIXME (aw): why is const auto& not possible for the following line?
+                // we only want cmds/vars from the required interface to be usable, not from it's child interfaces
+                std::string interface_name = req_route["required_interface"].get<std::string>();
+                auto requirement_impl_intf =
+                    config->get_interface_definition(interface_name);
```
a-w50, line 399: what's the difference to the auto requirement_impl_intf = config->get_interfaces()[requirement_module_name][requirement_impl_id] from before?

src/main.cpp
```diff
@@ -143,11 +143,12 @@ int main(int argc, char* argv[]) {
...
-            boost::dll::import_alias<void(std::function<Result(const std::string&, const std::string&, Parameters)>)>(
-                path, "everest_register_call_cmd_callback");
+            boost::dll::import_alias<void(std::function<Result(const std::string&, const uint64_t,
+                const std::string&, Parameters)>)>(path, "everest_register_call_cmd_callback");
 
-        auto call_cmd = [&everest](const std::string& param1, const std::string& param2, Parameters param3) {
-            return everest.call_cmd(param1, param2, param3);
+        auto call_cmd = [&everest](const std::string& requirement_id, const uint64_t idx,
+                                   const std::string& cmd_name, Parameters args) {
+            return everest.call_cmd(requirement_id, idx, cmd_name, args);
```
a-w50, line 151: :bowtie: for renaming params

include/framework/everest.hpp
```diff
@@ -89,8 +88,8 @@ class Everest {
     /// \brief Subscribes to a variable of another module identified by the given \p requirement_id and variable name \p
     /// var_name. The given \p callback is called when a new value becomes available
     ///
-    void subscribe_var(const std::string& requirement_id, const std::string& var_name, const ValueCallback& callback);
-    void subscribe_var(const std::string& requirement_id, const std::string& var_name, const JsonCallback& callback);
+    void subscribe_var(const std::string& requirement_id, uint64_t index, const std::string& var_name,
+                       const ValueCallback& callback);
```
a-w50, line 92: I wonder if we need these 2 subscribe_var overloads at all? The one with the ValueCallback should be sufficient I guess, because we shouldn't pass raw json - and if so, the raw json can also be put into boost::any

lib/config.cpp
```diff
@@ -65,26 +65,25 @@ static json parse_config_map(const json& config_map_schema, const json& config_m
...
-
-            json_validator validator(Config::loader, Config::format_checker);
-            validator.set_root_schema(config_entry);
-
-            try {
-                validator.validate(config_entry_value);
-            } catch (const std::exception& err) {
-                throw ConfigParseException(ConfigParseException::SCHEMA, config_entry_name, err.what());
+        json config_entry_value = config_map[config_entry_name];        // will be null if not present in json
+        if (!config_map.contains(config_entry_name) && config_entry.contains("default")) {
```
a-w50, line 81: instead of !config_map.contains(config_entry_name), I would prefer to check for config_entry_value == null

lib/config.cpp
```diff
@@ -65,26 +65,25 @@ static json parse_config_map(const json& config_map_schema, const json& config_m
...
+            config_entry_value = config_entry["default"];               // use default value defined in manifest
```
a-w50, line 82: Do we need this at all, because below we're using the patch

lib/config.cpp
```diff
@@ -164,32 +167,32 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
 
             json_validator validator(Config::loader, Config::format_checker);
             validator.set_root_schema(this->_schemas.manifest);
-            validator.validate(this->manifests[module_name]);
+            auto patch = validator.validate(this->manifests[module_name]);
+            if (!patch.is_null()) {
+                // extend manifest with default values
+                this->manifests[module_name] = this->manifests[module_name].patch(patch);
+            }
```
a-w50, line 181: It's good to have these patching process in all places, I wonder if this will change/break something because we haven't had it before.  There is no strict need for it, right?  For future I definitely would like to keep it

lib/config.cpp
```diff
@@ -164,32 +167,32 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
...
         // validate user-defined default values for the config meta-schemas
-        if (this->manifests[module_name].contains("config")) {
-            try {
-                validate_config_schema(this->manifests[module_name]["config"]);
-            } catch (const std::exception& e) {
-                EVLOG_AND_THROW(EVEXCEPTION(EverestConfigError,
-                                            "Failed to validate the module configuration meta-schema for module '",
-                                            module_name, "'. Reason:\n", e.what()));
-            }
+        try {
```
a-w50, line 187: Why did you drop the  if (this->manifests[module_name].contains("config"))?
It is not required to be existent, and below it is accessed, so this might lead to a null reference

lib/config.cpp
```diff
@@ -232,13 +234,7 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
...
                          << " against json schemas defined in module mainfest...";
 
-            // default value for missing implementation configs is an empty dict
-            json config_map = json({});
-            if (module_config.contains("config_implementation") &&
-                module_config["config_implementation"].contains(impl_id)) {
-                config_map = module_config["config_implementation"][impl_id];
-            }
-
+            json config_map = module_config["config_implementation"][impl_id];
```
a-w50, line 244: Probably again a null reference

lib/config.cpp
```diff
@@ -265,8 +261,7 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
 
         // validate config for !module
         {
-            // default value for missing !module config is an empty dict
-            json config_map = module_config.value("config_module", json({}));
+            json config_map = module_config["config_module"];
```
a-w50, line 271: is null the same as json({})? in nlohman json terms?

lib/everest.cpp
```diff
@@ -334,7 +333,7 @@ void Everest::publish_var(const std::string& impl_id, const std::string& var_nam
             oss << "Publish var of " << this->config.printable_identifier(this->module_id, impl_id)
```
a-w50, line 333: I really would like to change this error logging pattern with stringstream and oss, would be nice if this can be implemented with libfmt

lib/config.cpp
a-w50: Ah probably because of the patch which populates it with a default value of {}, nice!

lib/config.cpp
a-w50: if populated by default, then it's fine

schemas/manifest.json
```diff
@@ -2,33 +2,6 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "description": "Json schema for EVerest manifest files",
     "$defs": {
-        "varlist_subschema": {
```
a-w50, line 5: Good cleanup, didn't realize that this was still here

schemas/interface.json
```diff
@@ -23,7 +23,8 @@
                         "description": {
                             "type": "string",
                             "minLength": 2
-                        }
+                        },
+                        "default": false            // we don't allow default values
```
a-w50, line 49: Good point!
Btw, what happens if there is a default in the sub properties, is this already handled by the validate/patch mechanism?

lib/config.cpp
```diff
@@ -363,9 +370,28 @@ json Config::load_interface_file(const std::string& intf_name) {
 
         json interface_json = json::parse(intf_file);
 
+        // this subschema can not use allOf with the draft-07 schema because that will cause our validator to
+        // add all draft-07 default values which never validate (the {"not": true} default contradicts everything)
+        // --> validating against draft-07 will be done in an extra step below
```
a-w50, line 375: I don't really get this comment

lib/config.cpp
```diff
@@ -363,9 +370,28 @@ json Config::load_interface_file(const std::string& intf_name) {
...
+        validator.set_root_schema(draft07);
+        for (auto& var_entry : interface_json["vars"].items()) {
+            validator.validate(var_entry.value());
+        }
+        for (auto& cmd_entry : interface_json["cmds"].items()) {
+            for (auto& arguments_entry : interface_json["cmds"][cmd_entry.key()]["arguments"].items()) {
+                validator.validate(arguments_entry.value());
+            }
+            validator.validate(interface_json["cmds"][cmd_entry.key()]["result"]);
+        }
```
a-w50, line 394: :+1: for this splitup, it makes the logic clearer and keeps the json schema more readable

schemas/interface.json
```diff
@@ -2,12 +2,34 @@
...
                 },
-                {
+                "default": false            // we don't allow default values
+            },
+            "additionalProperties": true    // allow all sorts of entries (should be a json schema!)
+        },
+        "cmd_arguments_subschema": {
+            "type": "object",
+            "patternProperties": {
+                "^[a-zA-Z_][a-zA-Z0-9_.-]*$": { // arbitrary argument name
```
a-w50, line 32: Why isn't the type restriction to array and string not here?
Would it be possible to use the same meta subschema for var, cmd_arguments and cmd_result?

everestjs/everestjs.cpp
```diff
@@ -93,8 +93,7 @@ static Napi::Value setup_cmd_handler(const std::string& impl_id, const std::stri
                     std::vector<napi_value> args{ctx->cmd_handlers[cmd_key].Value(), ctx->js_module_ref.Value(), arg};
                     return args;
                 },
-                [&result](const Napi::Value& retval, bool) {
-                    // FIXME (aw): we could check the bool for failure ...
+                [&result](const Napi::Value& retval) {
```
a-w50, line 96: I'm not quite sure if this is the right way to go.  Because the on_reject isn't set now anymore, we have no chance to set a custom rejection handler.  I guess now we will only receive an exception, which we can't directly identify, i.e. from which async callback it came from - this would have been possible if we registered custom logic in the on_reject function.
So why we don't check the bool here? If it is true, we know that we're in the on_reject branch and retval should be our error object, which we can handle (that's why there was the FIXME comment) or just throw it right now.

a-w50 (changes requested):
I really appreciate all the work and improvements.  Basically I only have concerns with the "Don't handle rejections, terminate with a backtrace" commit.
Concerning the list handling, I think it's an elegant solution to put all connections, whether they are single ones or multiple into a vector.  This gives a bit overhead to the single connections, as they are treated now as a vector, but I don't think this will hurt performance.  We might also use a std::variant<connection, vector<connection>> for handling this - but for now the solution is just good simple.
I also appreciate the json schema related changes!
:+1:

include/framework/everest.hpp
hikinggrass: I think this might be something we should revisit during a cleanup after implementing type-support.

lib/config.cpp
hikinggrass: It refers to a (slightly surprising) behavior of the validator that would populate the patch with these draft-7 default values. Which in turn would lead to some very wrong non-validatable output  that we encountered during testing. Might be useful for future reference, but maybe we should clarify the intent a bit more.

a-w50: which default values, if I have a non-required integer property and will validate against draft7 with patch, will this lead to the property been set to zero then, or what?

hikinggrass: I have an example from the **enable command** of the **board_support_AC** interface which we found during testing:
In the main branch the validator would create the following output:
```json
{"description":"True: enabled, false: disabled.","type":"boolean"}
```

Without the changes mentioned in the comment we encountered the following output after validating with draft-7 and using the patch from the validation.
```json
{"additionalItems":true,"additionalProperties":true,"contains":true,"description":"True: enabled, false: disabled.","else":true,"if":true,"items":true,"not":true,"propertyNames":true,"readOnly":false,"then":true,"type":"boolean","uniqueItems":false}
```
The way it's implemented now prevents these (obviously wrong) additions to be patched in.

everestjs/utils.hpp
tmolitor-stud-tu: It is to get all the boost related error information (the current scope, file, line etc., basically everything you see in the log)

lib/config.cpp
tmolitor-stud-tu: just a copy/paste hiccup, may as well be used, yes.

lib/config.cpp
tmolitor-stud-tu: sure

lib/config.cpp
tmolitor-stud-tu: sure

schemas/interface.json
tmolitor-stud-tu: good question, I guess the current validation process (which includes patching the json with default values) will add defaults if not present in the data, so yes, it will be handled,but I don't know if that is desired behavior.

The better solution would be to remove all entries called `default` when loading the interface definition file, no matter their nesting level.

lib/config.cpp
tmolitor-stud-tu: What kai said. Even worse: these additions are contradicting in itself: you never will be able to validate something having `{"not": true}` in the schema no matter what you try to validate.

lib/config.cpp
tmolitor-stud-tu: and it prevents the "bug" discussed in the comment above :)

schemas/interface.json
tmolitor-stud-tu: not really, at leas for the upcoming default arguments stuff already in my pipeline.
I figured having the subschemas all separated makes it easier to update them in the future and also seems a bit cleaner.

everestjs/everestjs.cpp
tmolitor-stud-tu: no, to the contrary: we now have a complete js backtrace pointing us directly to the line of code that caused the exception.

tmolitor-stud-tu: Second: why catch it and rethrow it manually when we could leave it to the nodejs to do exactly that?

tmolitor-stud-tu:
I'm not sure if I can incorporate all these changes in time (e.g. in the next 1-2 weeks)

everestjs/everestjs.cpp
a-w50: But where would/could you catch this exception now?  If you want to have the backtrace, you could have just written
if (bool == true) throw Napi::Error
or something similar.  But I would leave it up to the user to decide on how to deal with rejected promises when using the JsExecContext helper class

a-w50:
This PR was merged via another branch

PR #6, opened 2022-01-14
Fixed location of include NodeApiVersion.cmake

hikinggrass:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ endif()
 # configure conan cmake, please consider adding dependencies to conanfile.txt
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
 list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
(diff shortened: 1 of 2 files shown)


a-w50 (changes requested):
Couldn't we use CMAKE_MODULE_PATH instead of this unhandy long path?

hikinggrass:
> Couldn't we use CMAKE_MODULE_PATH instead of this unhandy long path?

I think we already set the CMAKE_MODULE_PATH to "${CMAKE_SOURCE_DIR}/cmake" in the root CMakeLists.txt, but this doesn't seem to work

a-w50:
to CMAKE_SOURCE_DIR, here it is PROJECT_SOURCE_DIR

a-w50 (approved):
I hope it's the right way to choose PROJECT_SOURCE_DIR over CMAKE_SOURCE_DIR, but in case this project gets included via an add_subdirectory, this should be the way to go ...

corneliusclaussen (approved):

PR #7, opened 2022-01-19
Add more expressive error messages to call_cmd and publish_var json schema validation

hikinggrass:


lib/everest.cpp
```diff
@@ -181,8 +181,21 @@ json Everest::call_cmd(const std::string& requirement_id, const std::string& cmd
...
+            try {
+                validator.set_root_schema(cmd_definition["arguments"][arg_name]);
+                validator.validate(json_args[arg_name]);
+            } catch (const std::exception& e) {
+                std::ostringstream oss;
+                oss << "Call to "
+                    << this->config.printable_identifier(connection["module_id"], connection["implementation_id"])
+                    << "->" << cmd_name << "(";
+                for (auto const& key : arg_names) {
+                    oss << key << ",";
```
a-w50, line 193: this will leave an additional ',' at the end, probably we should refactor this exception handling code into separate functions

a-w50 (approved):
Good to have

PR #8, opened 2022-02-02
Adding support for ModuleInfo

a-w50:
- added new ModuleInfo type for holding information about a module
- added get_module_info function to retreive the data from the main
  config
- forwarding this information to the module init function
- changed get_manifests(): it now returns a const ref to the json dict


hikinggrass (approved):
Looks good

PR #9, opened 2022-02-02
Pretty print json in validation error message

corneliusclaussen:
Signed-off-by: Cornelius Claussen <cc@pionix.de>

### DIFF
--- a/lib/everest.cpp
+++ b/lib/everest.cpp
@@ -332,7 +332,7 @@ void Everest::publish_var(const std::string& impl_id, const std::string& var_nam
         } catch (const std::exception& e) {
             std::ostringstream oss;
             oss << "Publish var of " << this->config.printable_identifier(this->module_id, impl_id)
-                << " with variable name '" << var_name << "' with value '" << json_value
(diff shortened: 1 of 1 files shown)


lib/everest.cpp
```diff
@@ -332,7 +332,7 @@ void Everest::publish_var(const std::string& impl_id, const std::string& var_nam
         } catch (const std::exception& e) {
             std::ostringstream oss;
             oss << "Publish var of " << this->config.printable_identifier(this->module_id, impl_id)
-                << " with variable name '" << var_name << "' with value '" << json_value
+                << " with variable name '" << var_name << "' with value '" << std::setw(4) << json_value
```
hikinggrass, line 335: You could also use
```cpp
json_value.dump(4)
```
here which [would indent the whole json properly.](https://json.nlohmann.me/api/basic_json/dump/)

corneliusclaussen: I think this is identical

hikinggrass: Thats true.

hikinggrass (approved):

PR #10, opened 2022-02-10
Update error handling and logging

a-w50:
- instead of silencing completely, everestjs throws now errors on
  unhandled rejections, which in turn will lead again to an unhandled
  rejection within the callbackWrapper and could thus yield a stack trace
  (this seems to depend on nodejs version and its runtime configuration,
  because an unhandled rejection doesn't seem to be the same like an
  error)
- append boost log information to napi error exceptions
- use EVLOG_AND_THROW
- added some try and catch blocks, where necessary

Signed-off-by: Thilo Molitor <thilo@eightysoft.de>
Co-authored-by: aw <aw@pionix.de>


a-w50:
This is split out of the requirement_lists PR, which is not really related to it

hikinggrass (approved):
Looks good!

PR #11, opened 2022-02-10
Feature/requirement lists ng

a-w50:
Refactored, edited and cleaned changes from wip/requirement_lists

### DIFF
--- a/everestjs/everestjs.cpp
+++ b/everestjs/everestjs.cpp
@@ -41,7 +41,7 @@ struct EvModCtx {
 
     std::unique_ptr<JsExecCtx> js_cb;
 
-    std::map<std::pair<std::string, std::string>, Napi::FunctionReference> var_subscriptions;
+    std::map<std::pair<Requirement, std::string>, Napi::FunctionReference> var_subscriptions;
     std::map<std::pair<std::string, std::string>, Napi::FunctionReference> cmd_handlers;
 
     std::map<std::string, Napi::FunctionReference> mqtt_subscriptions;
@@ -112,7 +112,7 @@ static Napi::Value setup_cmd_handler(const std::string& impl_id, const std::stri
     return env.Undefined();
 }
 
-static Napi::Value set_var_subscription_handler(const std::string& requirement_id, const std::string& var_name,
+static Napi::Value set_var_subscription_handler(const Requirement& req, const std::string& var_name,
                                                 const Napi::CallbackInfo& info) {
     BOOST_LOG_FUNCTION();
     const auto& env = info.Env();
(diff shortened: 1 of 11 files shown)


tmolitor-stud-tu:
why squashing this into only two commits?
by keeping the smaller commits its really easy to revert one of the changes (for example the "don'T allow default values thing), when squasing everything into one or two commits, undoing individual changes is much more complicated.

that's a policy thing and thus that's a question for all
@hikinggrass 
@a-w50 
@caller 
@corneliusclaussen 
@pionix-juraed

corneliusclaussen:
We decided to have the main history contain larger blocks instead of fine grained commits

everestjs/everestjs.cpp
```diff
@@ -381,93 +380,99 @@ static Napi::Value boot_module(const Napi::CallbackInfo& info) {
...
+            // if this was a requirement with min_connections == 1 and max_connections == 1,
+            // this will be simply a single connection, but an array of connections otherwise
+            // (this array can have only one entry, if only one connection was provided, though)
+            const bool is_list = req_route_list.is_array();
+            if (!is_list) {
+                req_route_list = json::array({req_route_list});
             }
+            auto req_mod_vars_array = Napi::Array::New(env);
+            auto req_mod_cmds_array = Napi::Array::New(env);
+            for (uint64_t i = 0; i < req_route_list.size(); i++) {
```
hikinggrass: I changed this to uint64_t because we use this type for the Requirement index

include/utils/types.hpp
```diff
@@ -26,19 +26,35 @@ using ValueCallback = std::function<void(Value)>;
...
 
+struct ModuleInfo {
+    std::string name;
+    std::vector<std::string> authors;
+    std::string license;
+    std::string id;
+};
+
+struct Requirement {
+    Requirement(const std::string& requirement_id, uint64_t index) : id(requirement_id), index(index){};
```
hikinggrass: Changed index to uint64_t to prevent narrowing conversion

include/utils/types.hpp
```diff
@@ -26,19 +26,35 @@ using ValueCallback = std::function<void(Value)>;
...
+        if (id < rhs.id) {
+            return true;
+        } else if (id == rhs.id) {
+            return (index < rhs.index);
+        } else {
+            return false;
+        }
+    }
+    const std::string id;
+    uint64_t index;
```
hikinggrass: Changed index to uint64_t to prevent narrowing conversion

hikinggrass (approved):
I had to change the index of Requirement to uint64_t to prevent narrowing conversion, or was that a conscious decision  to use int here? Then we have to reflect that in the ev-cli code generator as well.

everestjs/everestjs.cpp
tmolitor-stud-tu: No, no concious decisision, uint64_t is fine

tmolitor-stud-tu:
> We decided to have the main history contain larger blocks instead of fine grained commits

I think thats a bad decision for various reasons (bisect, document what change has which effect etc.)

PR #12, opened 2022-02-15
Properly check for empty return types

a-w50:
- this led to segmentation faults for cmds which don't return
  anything

### DIFF
--- a/lib/everest.cpp
+++ b/lib/everest.cpp
@@ -760,13 +760,22 @@ std::string Everest::check_args(const Arguments& func_args, json manifest_args) 
 bool Everest::check_arg(ArgumentType arg_types, json manifest_arg) {
     BOOST_LOG_FUNCTION();
 
-    std::ostringstream oss;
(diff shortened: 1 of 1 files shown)


lib/everest.cpp
```diff
@@ -760,9 +760,14 @@ std::string Everest::check_args(const Arguments& func_args, json manifest_args)
 bool Everest::check_arg(ArgumentType arg_types, json manifest_arg) {
     BOOST_LOG_FUNCTION();
 
+    // FIXME (aw): what about this oss? Nobody uses it
```
hikinggrass: That was probably intended to be able to construct the error message that should be logged before returning from this function.
However since EVLOG takes a stream we can just use it directly, and fix another bug in the process: that the error messages are only written to oss, but never logged.

hikinggrass (changes requested):

hikinggrass (approved):

PR #14, opened 2022-02-16
Introducing libfmt for string formatting

a-w50:

### DIFF
--- a/lib/config.cpp
+++ b/lib/config.cpp
@@ -5,6 +5,8 @@
 #include <set>
 #include <sstream>
 
+#include <fmt/format.h>
+
 #include <everest/exceptions.hpp>
 #include <everest/logging.hpp>
 
@@ -47,7 +49,7 @@ static void validate_config_schema(const json& config_map_schema) {
             validator.set_root_schema(config_item.value());
             validator.validate(config_item.value()["default"]);
         } catch (const std::exception& e) {
-            throw std::runtime_error("Config item '" + config_item.key() + "' has issues:\n" + e.what());
+            throw std::runtime_error(fmt::format("Config item '{}' has issues:\n{}", config_item.key(), e.what()));
         }
     }
 }
@@ -117,7 +119,7 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
     fs::path config_path = config_file;
 
     try {
-        EVLOG(info) << "Loading config file at: " << fs::canonical(config_path).c_str();
+        EVLOG(info) << fmt::format("Loading config file at: {}", fs::canonical(config_path).string());
 
         std::ifstream ifs(config_path.c_str());
         std::string config_file((std::istreambuf_iterator<char>(ifs)), (std::istreambuf_iterator<char>()));
@@ -130,7 +132,7 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
         // TODO(kai): or should we introduce a "meta-config" that references all configs that should be merged here?
         auto user_config_path = config_path.parent_path() / "user-config" / config_path.filename();
         if (fs::exists(user_config_path)) {
-            EVLOG(info) << "Loading user-config file at: " << fs::canonical(user_config_path).c_str();
+            EVLOG(info) << fmt::format("Loading user-config file at: {}", fs::canonical(user_config_path).string());
 
             std::ifstream ifs(user_config_path.c_str());
             std::string user_config_file((std::istreambuf_iterator<char>(ifs)), (std::istreambuf_iterator<char>()));
@@ -151,7 +153,7 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
             this->main = this->main.patch(patch);
         }
     } catch (const std::exception& e) {
-        EVLOG_AND_THROW(EVEXCEPTION(EverestConfigError, "Failed to load and parse config file: ", e.what()));
+        EVLOG_AND_THROW(EverestConfigError(fmt::format("Failed to load and parse config file: {}", e.what())));
     }
 
     // load manifest files of configured modules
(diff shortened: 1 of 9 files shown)


lib/everest.cpp
```diff
@@ -233,13 +206,12 @@ json Everest::call_cmd(const Requirement& req, const std::string& cmd_name, json
...
         if (data["id"] != call_id) {
-            EVLOG(debug) << "RES: data_id != call_id: '" << data["id"] << " != " << call_id;
+            EVLOG(debug) << fmt::format("RES: data_id != call_id ({} != {})", data["id"], call_id);
             return;
         }
 
-        EVLOG(debug) << "Incoming res " << data["id"] << " for "
-                     << this->config.printable_identifier(connection["module_id"], connection["implementation_id"])
-                     << "->" << cmd_name << "()";
+        EVLOG(debug) << "Incoming res {} for {}->{}()", data["id"],
```
LAD101work: This seems like there is a call to `fmt::format()` missing. Please check!

lib/everest.cpp
```diff
@@ -517,12 +484,9 @@ void Everest::provide_cmd(const std::string impl_id, const std::string cmd_name,
         }
 
         std::ostringstream oss;
```
LAD101work: Is this stream object still needed? I cannot find any usage.

lib/config.cpp
```diff
@@ -533,7 +536,7 @@ schemas Config::load_schemas(std::string schemas_dir) {
     BOOST_LOG_FUNCTION();
     schemas schemas;
 
-    EVLOG(info) << "Loading base schema files for config and manifests... from:" << schemas_dir;
+    EVLOG(info) << fmt::format("Loading base schema files for config and manifests... from:", schemas_dir);
```
LAD101work: Not sure if this might work or not, but it surely would be more clear to put the replacement operators `{}` in at the end of the string.
Please check!

LAD101work (changes requested):
Please check comments.

lib/config.cpp
a-w50: :+1: for your findings! Thanks!!

LAD101work: You're welcome!

lib/config.cpp
```diff
@@ -466,6 +468,7 @@ ModuleConfigs Config::get_module_configs(const std::string& module_id) {
                     value = data.get<double>();
                 } else if (data.is_number_integer()) {
                     if (entry_type == "number") {
+                        // FIXME (aw): where does this belong to?
```
hikinggrass: This belonged to debug output that printed every config entry like this:
```cpp
EVLOG(debug) << " config entry: " << entry.key();
```
However, the corresponding debug output has since been removed. So I would suggest we remove this or reword it,

lib/everest.cpp
```diff
@@ -328,11 +299,10 @@ void Everest::publish_var(const std::string& impl_id, const std::string& var_nam
             validator.set_root_schema(var_definition);
             validator.validate(json_value);
         } catch (const std::exception& e) {
-            std::ostringstream oss;
-            oss << "Publish var of " << this->config.printable_identifier(this->module_id, impl_id)
-                << " with variable name '" << var_name << "' with value '" << std::setw(4) << json_value
-                << "' could not be validated with schema: " << e.what();
-            EVLOG_AND_THROW(EverestApiError(oss.str()));
+            // FIXME (aw): previously there was an setw(4) for output of the json_value here, is that needed?
```
hikinggrass: That would be needed to pretty-print the json value.
However, you can use json_value.dump(4) here, [that is the non-stream equivalent.](https://stackoverflow.com/a/47835694)

lib/everest.cpp
```diff
@@ -516,13 +484,9 @@ void Everest::provide_cmd(const std::string impl_id, const std::string cmd_name,
...
-        std::ostringstream oss;
-        oss << "Incoming " << this->config.printable_identifier(this->module_id, impl_id) << "->" << cmd_name << "(";
-        for (auto const& key : arg_names) {
-            oss << key << ",";
-        }
-        oss << ") for <handler>";
-        EVLOG(debug) << oss.str();
+        EVLOG(debug) << fmt::format("Incoming {}->{}({}) for <handler>",
+                                    this->config.printable_identifier(this->module_id, impl_id), cmd_name,
+                                    fmt::join(arg_names, ","));
```
hikinggrass, line 488: I really like this fmt::join syntax

lib/everest.cpp
```diff
@@ -575,23 +540,22 @@ void Everest::provide_cmd(const std::string impl_id, const std::string cmd_name,
                 }
 
             } catch (const std::exception& e) {
-                EVLOG(warning) << "Ignoring return value of cmd '" << cmd_name
-                               << "' because the validation of the result failed. " << e.what();
-                EVLOG(warning) << "definition: " << cmd_definition;
-                EVLOG(warning) << "data: " << res_publish_data;
+                EVLOG(warning) << fmt::format("Ignoring return value of cmd '{}' because the validation of the result "
```
hikinggrass, line 542: Good refactor

lib/mqtt_abstraction_impl.cpp
```diff
@@ -250,19 +254,14 @@ void MQTTAbstractionImpl::unregister_handler(const std::string& topic, const Tok
...
-    std::ostringstream oss;
-    oss << "#handler[" << topic << "] = ";
-    if (this->handlers.count(topic) == 0) {
-        oss << "None";
-    } else {
-        oss << this->handlers[topic].size();
-    }
-    EVLOG(debug) << oss.str();
+    const std::string handler_count =
+        (this->handlers.count(topic) == 0) ? "None" : std::to_string(this->handlers[topic].size());
```
hikinggrass, line 263: Usually I'm not the biggest fan of this short if syntax, but here I would say it's acceptable.

hikinggrass (changes requested):
I like it, I think we can merge this with just a few minor changes.

lib/everest.cpp
a-w50: Ok, but I see no reason here to have a minimum width of 4, if the value is anyway put inside the ' ' pair inline. So I just leave it as it is

lib/config.cpp
a-w50: yes, I will remove it.  It should be just valid to have config entries of type number, which are set as integers

lib/mqtt_abstraction_impl.cpp
a-w50: From a performance perspective it is the right thing to do, but from the readability - it's ok ...
one could write a lambda or helper function for it to make it more readable, but on the other hand it is just that simple and used only once ..

hikinggrass (approved):

lib/everest.cpp
corneliusclaussen: If no min width is set the json is dumped unformatted and the debug output is essentially unusable.

a-w50: Ok, I got it now.  I thought setw(4) will only lead to minimum with of 4 for the output of the value. But it seems that the nlomann stream operator interpretes this value differently as a indention width.  Furthermore if it is printed over many lines, the value shouldn't be inlined directly into the formatting string, but rather newline separated

LAD101work (approved):
Very good! Nice and clean now!

lib/everest.cpp
LAD101work: Resolved!

lib/everest.cpp
LAD101work: Resolved!

lib/config.cpp
LAD101work: Resolved!

PR #16, opened 2022-02-21
Add support for complex types defined in json files.

hikinggrass:
These files reside in the types folder, next to the interfaces folder.


include/utils/config.hpp
```diff
@@ -25,8 +25,11 @@ struct schemas {
     json config;    ///< The config schema
     json manifest;  ///< The manifest scheme
     json interface; ///< The interface schema
+    json type;      ///< The type schema
 };
 
+const static std::regex type_uri_regex{R"(^((?:\/[a-zA-Z0-9\-\_]+)+#\/[a-zA-Z0-9\-\_]+)$)"};
+
```
a-w50, line 35: A comment or link to what this exactly does would be nice

lib/config.cpp
```diff
@@ -554,13 +582,38 @@ void Config::loader(const json_uri& uri, json& schema) {
...
+
+    if (uri.location() == "http://json-schema.org/draft-07/schema") {
+        schema = nlohmann::json_schema::draft7_schema_builtin;
+        return;
+    } else {
+        auto path = uri.path();
+        if (this->types.contains(path)) {
+            schema = this->types[path];
+            EVLOG_verbose << fmt::format("ref path \"{}\" schema has been found.", path);
+            return;
```
a-w50, line 596: Is there any order necessary for loading the the type files? For example yeti types depend on unit types.  What if yeti.json gets loaded first?

lib/conversions.cpp
```diff
@@ -98,7 +98,8 @@ template <> json convertTo<json>(Parameters params) {
         } else if (param.second.type() == typeid(Object)) {
             j[param.first] = boost::any_cast<Object>(param.second);
         } else {
-            EVLOG_AND_THROW(EverestApiError(fmt::format("WRONG C++ TYPE: {}", param.second.type().name()))); // FIXME
+            EVLOG_AND_THROW(EverestApiError(
+                fmt::format("WRONG C++ TYPE: {} FOR ENTRY: {}", param.second.type().name(), param.first))); // FIXME
```
a-w50: The uppercase should be removed

lib/everest.cpp
```diff
@@ -188,7 +188,8 @@ json Everest::call_cmd(const Requirement& req, const std::string& cmd_name, json
     }
 
     if (this->validate_data_with_schema) {
-        json_validator validator(Config::loader, Config::format_checker);
+        json_validator validator([this](const json_uri& uri, json& schema) { this->config.ref_loader(uri, schema); },
+                                 Config::format_checker);
```
a-w50: I think it's also possible to extend the json_validator and use that as the validator object here.  Might be cleaner and more flexible then the lambda here

lib/everest.cpp
```diff
@@ -533,7 +539,9 @@ void Everest::provide_cmd(const std::string impl_id, const std::string cmd_name,
                 // only use validator on non-null return types
                 if (!(res_data["retval"].is_null() &&
                       (!cmd_definition.contains("result") || cmd_definition["result"].is_null()))) {
-                    json_validator validator(Config::loader, Config::format_checker);
+                    json_validator validator(
+                        [this](const json_uri& uri, json& schema) { this->config.ref_loader(uri, schema); },
+                        Config::format_checker);
```
a-w50, line 545: Yep, this validator object gets created 5 times with this "complex" construction.  This should be refactored

a-w50 (changes requested):

lib/config.cpp
hikinggrass: That should not be an issue here since all type json files have been loaded at startup into the this->types member

lib/conversions.cpp
```diff
@@ -31,7 +31,7 @@ template <> json convertTo<json>(Result retval) {
         } else if (ret_any.type() == typeid(Object)) {
             return boost::any_cast<Object>(ret_any);
         } else {
-            EVLOG_AND_THROW(EverestApiError(fmt::format("WRONG C++ TYPE: {}", ret_any.type().name()))); // FIXME
+            EVLOG_AND_THROW(EverestApiError(fmt::format("Wrong C++ typ: {}", ret_any.type().name()))); // FIXME
```
a-w50: Typo in typ :rofl:

a-w50 (approved):
Alright

PR #17, opened 2022-03-03
Dump json variables before calling fmt::format

hikinggrass:
You have to call dump() on a json variable before fmt::format-ting it.

Otherwise this might lead to very obscure exceptions like:
```
Dynamic exception type: nlohmann::detail::type_error
std::exception::what: [json.exception.type_error.302] type must be string, but is boolean
terminate called after throwing an instance of 'nlohmann::detail::type_error'
  what():  [json.exception.type_error.302] type must be string, but is boolean
```


a-w50 (approved):
Well done!

PR #22, opened 2022-03-24
Improve MQTT message handling

hikinggrass:
- Support for MQTT Quality of Service
- Removal of now superfluous "ack" mechanism for commands
- Add message queues to make sure variables and commands are delivered in the order they were received via MQTT
- Additionally merge all command and variable topics of a (sub)module into a "cmd" and  "var" topic to make sure there are not raceconditions between topics
- Use of shorter MQTT prefixes consisting only of the (sub)module name
- To make sure no information about the loaded modules is lost publish metadata containing information about the module type and its provided interfaces via MQTT
- Cache messages that are published before a connection to the MQTT broker is established and send them then
- Various smaller cleanups

### DIFF
--- a/lib/message_queue.cpp
+++ b/lib/message_queue.cpp
@@ -1,0 +1,112 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+#include <thread>
+
+#include <fmt/format.h>
+
+#include <everest/logging.hpp>
+
+#include <utils/message_queue.hpp>
+
+namespace Everest {
+
+MessageQueue::MessageQueue(const std::function<void(std::shared_ptr<Message> message)>& message_callback) :
+    message_callback(message_callback), running(true) {
+    this->worker_thread = std::thread([this]() {
+        while (this->running) {
+            std::shared_ptr<Message> message;
+            std::unique_lock<std::mutex> lock(this->message_mutex);
+            this->cv.wait(lock, [this]() { return !this->message_queue.empty(); });
+
+            message = this->message_queue.front();
+            this->message_queue.pop();
+            lock.unlock();
+
+            // pass the message to the message callback
+            this->message_callback(message);
+        }
+    });
+}
+
+void MessageQueue::add(std::shared_ptr<Message> message) {
+    {
+        std::lock_guard<std::mutex> lock(this->message_mutex);
+        this->message_queue.push(message);
+    }
+    this->cv.notify_all();
+}
+
+void MessageQueue::stop() {
+    this->running = false;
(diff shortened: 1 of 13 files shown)


hikinggrass:
> One could probably define the QOS enum directly with the MQTT_PUBLISH_QOS defines, to safe the switch case On the other hand this would leak implementation details

Unfortunately you would also not be able to use the same QOS enum for publish and subscribe, given that the MQTT-C library uses the MQTT_PUBLISH_QOS_ flags for publishing, but just a simple integer for subscribing

a-w50 (changes requested):
See my comments.  I would request you to rework the shared_ptr handling.  Thank you for that pile of work

lib/message_queue.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+MessageQueue::MessageQueue(const std::function<void(std::shared_ptr<Message> message)>& message_callback) :
+    message_callback(message_callback), running(true) {
+    this->worker_thread = std::thread([this]() {
+        while (this->running) {
+            std::shared_ptr<Message> message;
+            std::unique_lock<std::mutex> lock(this->message_mutex);
+            this->cv.wait(lock, [this]() { return !this->message_queue.empty(); });
+
+            message = this->message_queue.front();
```
a-w50, line 24: you could use auto message = std::move(this->message_queue.front()) here, instead of initializing, copying and destroying it, but the pop should follow closeby

lib/message_queue.cpp
```diff
@@ -0,0 +1,142 @@
...
+            this->message_queue.pop();
+            lock.unlock();
+
+            auto data = *message.get();
+
+            // get the registered handlers
+            std::vector<std::shared_ptr<TypedHandler>> local_handlers;
```
a-w50, line 60: performance wise (although it doesn't matter here), it would be better to copy construct the vector directly instead of iterating and incrementing it one by one

a-w50 (approved):
Nice additions!
I'm still not convinced, that the MessageHandler.handlers needs to be a container of shared_ptr of TypeHandlers, and that the TypedHandler.handler needs to be a shared_ptr of a std::function.
Creating a shared_ptr (and its dereferencing) wont save you much, than just copying or referencing the std::function object.
For the MessageHandler.handlers its somehow convinient, because you can use the shared_ptr as an index (aka Token) into the set, which you can't do directly with the TypeHandler.  The TypeHandler could be referenced and moved.  I think it's not necessary to do right now, but it might increase performance and maybe readability
Thanks for the work!

a-w50:
I also tested building and running against the run_sil.sh from everest-framework (8a36cc596).
Happy squashing ...

PR #23, opened 2022-04-12
Correctly serialize output to satisfy nlohmann lib

LAD101work:
Added correct serialization to fmt output when displaying error messages from validator.

Signed-off-by: LAD101work <96466764+LAD101work@users.noreply.github.com>


hikinggrass (approved):

a-w50 (commented):
Good find, it's a pitty, that I forgot to check this formatted dumping when doing the switch to fmt.
If you're willing to, please create a global constant for the indentation width, so we could change it in one place and don't have the '2's scattered arround

PR #28, opened 2022-04-19
Move RuntimeSettings into framework, build modules as binaries

hikinggrass:
Modules can still be built and loaded as shared libraries by setting the CMake option BUILD_MODULE_SHARED_LIBRARIES to ON


CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ endif()
 
 option(CREATE_SYMLINKS "Create symlinks to javascript modules and auxillary files - for development purposes" OFF)
 
+option(BUILD_MODULE_SHARED_LIBRARIES "Build modules as shared libraries" OFF)
+
+option(BUILD_MODULE_BINARIES "Build modules as binaries" ON)
```
a-w50: Is there any reason why we should allow both options instead of xor?
Do shared modules have any advantage over the binaries?

include/framework/runtime.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+struct ModuleCallbacks {
+    std::function<void(ModuleAdapter::CallFunc call)> register_call_handler;
+    std::function<void(ModuleAdapter::PublishFunc publish)> register_publish_handler;
+    std::function<void(ModuleAdapter::SubscribeFunc subscribe)> register_subscribe_handler;
+    std::function<void(ModuleAdapter::ExtMqttPublishFunc ext_mqtt_publish)> register_ext_mqtt_publish_handler;
+    std::function<void(ModuleAdapter::ExtMqttSubscribeFunc ext_mqtt_subscribe)> register_ext_mqtt_subscribe_handler;
+    std::function<std::vector<cmd>(const json& connections)> everest_register;
+    std::function<void(ModuleConfigs module_configs, const ModuleInfo& info)> init;
+    std::function<void()> ready;
```
a-w50: Probably this commit would be a good point to remove all those different register_handler functions.  Except for the ext_mqtt functions we want to register always all the functions and not just one by one.  I would propose, that instead there will be only one single hook, which the module as a library (in case we'll keep it as a library) will offer - something like register_everest, and this single function will then receive a struct with all the necessary functions to access everest.  So instead of registering n single callbacks, one register with a struct of n callbacks.

lib/runtime.cpp
```diff
@@ -0,0 +1,228 @@
...
+    const std::function<void()>& ready) :
+    register_call_handler(register_call_handler),
+    register_publish_handler(register_publish_handler),
+    register_subscribe_handler(register_subscribe_handler),
+    register_ext_mqtt_publish_handler(register_ext_mqtt_publish_handler),
+    register_ext_mqtt_subscribe_handler(register_ext_mqtt_subscribe_handler),
+    everest_register(everest_register),
+    init(init),
+    ready(ready) {
+}
```
a-w50: This could be avoided by the above mentioned single register function

lib/runtime.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+        // register the modules ready handler with the framework
+        // this handler gets called when the global ready signal is received
+        everest.register_on_ready_handler(this->callbacks.ready);
+
+        // the module should now be ready
+        everest.signal_ready();
+
+        mainloop_thread.join();
+
```
a-w50: I guess there is a lot of code duplication here with main.cpp.  If we will drop main.cpp, then this doesn't matter.  If not, it would be nice to share common source.

lib/runtime.cpp
```diff
@@ -0,0 +1,228 @@
...
+    this->runtime_settings = std::make_unique<RuntimeSettings>(vm);
+    this->original_process_name = argv[0];
+
+    if (vm.count("module") != 0) {
+        this->module_id = vm["module"].as<std::string>();
+    } else {
+        EVTHROW(EVEXCEPTION(EverestApiError, "--module parameter is required"));
+    }
+
+    return true;
```
a-w50, line 209: Although this gives additional flexibility, I don't really like the fact, that the manager has already parsed all the arguments, just to forward them and let them parsed again.  Something related to this would also be, that the spawned modules shouldn't know about the complete configuration - shouldn't they only get their module and implementation specific config (which could be passed by the manager via mqtt?)

a-w50 (approved):
I'm happy with the conceptual implementation, although I think we could do more refactoring and unification.
The difference between the shared modules and the binary is in the end just a matter of linkage.  Concerning the capability stuff we could equally put the same main binary into each of the module path and let it load the shared library module.
But I would almost argue, that we don't need the library as a shared module, as long as we are using it in a separate process, because we win nothing using it - i.e. we want to have the capabilty feature and hence need a separate loader binary for every module, the shared module library doesn't get used by multiple processes, so no memory advantages, and lastly the dynamic loading is less performant than a statically linked binary (that might be still link dynamically against libeverest).

include/framework/runtime.hpp
hikinggrass: Sure, that sounds good. I'll do a little more refactoring in this series of PRs.

lib/runtime.cpp
hikinggrass: I think I'll drop the modules as shared libraries completely and remove the main.cpp in that process.

lib/runtime.cpp
hikinggrass: I like the idea of removing the double-parsing of the config a lot, but since this will probably be a bigger change I'll tackle that in a dedicated PR

PR #32, opened 2022-04-25
Access to Everest namespace can be ambiguous which fails in clang-14

hikinggrass:


LAD101work (approved):
This works! Very good!

PR #38, opened 2022-05-20
Add python interface

hikinggrass:
This allows modules to be written in Python.

Uses pybind11 for the python bindings and is based partially on the everestjs javascript bindings.


corneliusclaussen:
replaced by feature/python-bindings-compressed

PR #41, opened 2022-06-20
Use type hints for ambiguous integer/number types

hikinggrass:


corneliusclaussen (approved):
tested with number in interface/double in c++ and integer/int

PR #42, opened 2022-07-05
Removed obsolete EVEREST_CROSS_COMPILE option

hikinggrass:


a-w50 (approved):
Maybe we can add a simple comment above lines like add_definitions(-DBOOST_LOG_DYN_LINK), so one knows why they are there

PR #43, opened 2022-07-05
Add some common serial port and date conversion functionality directly to the framework

hikinggrass:
This includes open/close of the device, setting of common attributes and a COBS de/encoder as well as crc32 calculation

### DIFF
--- a/lib/date.cpp
+++ b/lib/date.cpp
@@ -5,11 +5,11 @@
 namespace Everest {
 namespace Date {
 
-std::string to_rfc3339(std::chrono::time_point<date::utc_clock> t) {
+std::string to_rfc3339(const std::chrono::time_point<date::utc_clock>& t) {
     return date::format("%FT%TZ", std::chrono::time_point_cast<std::chrono::milliseconds>(t));
 }
 
-std::chrono::time_point<date::utc_clock> from_rfc3339(std::string t) {
+std::chrono::time_point<date::utc_clock> from_rfc3339(const std::string& t) {
     std::istringstream infile{t};
     std::chrono::time_point<date::utc_clock> tp;
     infile >> date::parse("%FT%T", tp);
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -15,19 +15,7 @@ if (NOT CMAKE_BUILD_TYPE)
...
-    endif()
-    set(Boost_USE_STATIC_LIBS ON)
-    set(CMAKE_SYSTEM_NAME Linux)
-    set(CMAKE_SYSTEM_PROCESSOR arm)
-    set(BOOST_LOG_DYN_LINK OFF)
-    set(BOOST_ALL_DYN_LINK OFF)
-else()
-    add_definitions(-DBOOST_LOG_DYN_LINK)
-endif()
+add_definitions(-DBOOST_LOG_DYN_LINK)
```
a-w50: Note, this change is already part of the other PR

lib/date.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <utils/date.hpp>
+
+namespace Everest {
+namespace Date {
+
+std::string to_rfc3339(std::chrono::time_point<date::utc_clock> t) {
+    return date::format("%FT%TZ", std::chrono::time_point_cast<std::chrono::milliseconds>(t));
+}
+
+std::chrono::time_point<date::utc_clock> from_rfc3339(std::string t) {
```
a-w50: t should be const ref

lib/date.cpp
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+#include <utils/date.hpp>
+
+namespace Everest {
+namespace Date {
+
+std::string to_rfc3339(std::chrono::time_point<date::utc_clock> t) {
```
a-w50: t should be const ref

a-w50 (approved):
For now it is good, that the serial stuff can be reused.  But it needs to be refactored, especially the cobs/crc stuff (printf ...)

PR #45, opened 2022-07-14
Improve ready message

corneliusclaussen:
The current message is a bit confusing since "all modules ready" means that they have run the init function but not the ready function yet. So they are ready to run ready(). Also this message is a little easier to see to make it easier to find out when everest is actually usable on long startups.

Signed-off-by: Cornelius Claussen <cc@pionix.de>


a-w50 (changes requested):
Looks like you didn't format the code?
furthermore please use
fmt::format(TERMINAL_STYLE_OK, "your text"), instead of the hard-coded literals
there is also TERMINAL_STYLE_ERROR

a-w50 (approved):

PR #49, opened 2022-09-26
Ability to provide user config as parameter

hikinggrass:


include/utils/config.hpp
```diff
@@ -85,7 +85,7 @@ class Config {
     /// \brief creates a new Config object, looking for the config.json and schemes folder relative to the provided \p
     /// main_dir
     explicit Config(std::string schemas_dir, std::string config_file, std::string modules_dir,
-                    std::string interfaces_dir, std::string types_dir);
+                    std::string interfaces_dir, std::string types_dir, std::string user_config_file);
```
a-w50, line 88: Would be good practice here, to change these to const refs

lib/config.cpp
```diff
@@ -133,7 +133,7 @@ Config::Config(std::string schemas_dir, std::string config_file, std::string mod
         // config_file. The config is supposed to have the same name as the parent config.
         // TODO(kai): introduce a parameter that can overwrite the location of the user config?
         // TODO(kai): or should we introduce a "meta-config" that references all configs that should be merged here?
-        auto user_config_path = config_path.parent_path() / "user-config" / config_path.filename();
+        auto user_config_path = fs::path(user_config_file);
         if (fs::exists(user_config_path)) {
             EVLOG_debug << fmt::format("Loading user-config file at: {}", fs::canonical(user_config_path).string());
```
a-w50, line 139: I don't like the else handling here ...
The question is, why we're using weakly_canonical in the first place.  So if the file doesn't exists, the user_config_file will be still set to the value, a user might have supplied.  Now in case it doesn't exist it will silently fail and log "No user-config provided."  This is not what I would expect the program to behave

src/manager.cpp
```diff
@@ -587,7 +587,7 @@ int boot(const po::variables_map& vm) {
                 shutdown_modules(module_handles, *config, mqtt_abstraction);
                 modules_started = false;
 
-		// Exit if a module died, this gives systemd a change to restart manager
+                // Exit if a module died, this gives systemd a chance to restart manager
```
a-w50, line 590: :+1:

lib/runtime.cpp
```diff
@@ -42,13 +42,21 @@ RuntimeSettings::RuntimeSettings(const po::variables_map& vm) : main_dir(vm["mai
         config_file = main_dir / "conf/config.json";
     }
 
+    if (vm.count("user_conf")) {
+        user_config_file = vm["user_conf"].as<std::string>();
+    } else {
+        user_config_file = config_file.parent_path() / "user-config" / config_file.filename();
```
a-w50, line 48: Furthermore, if no user-config is supplied via command-line, it might still happen, that one gets loaded ...
This is some kind of default behavior, but it should be explicitly stated with log info

a-w50 (changes requested):

include/utils/config.hpp
hikinggrass: would probably be best to change it to (boost/std)::filesystem paths

lib/runtime.cpp
hikinggrass: The problem with stating this with a log info message here is that it will be printed for every module

lib/config.cpp
hikinggrass: Maybe it would be best to perform this initial check in the manager and differentiate there between the following:

- user config empty -> maybe log a debug message
- user config set to valid path -> log info message that this file will be used
- user config set to invalid path -> log error and halt

include/utils/config.hpp
a-w50: That's another point, but these should also be const refs :laughing: 
I'm really sure if these would need to be filesystem paths in case of the constructor, imagine one day you want to support something like config_file="http;//link.to/config_file"
maybe we should keep them as strings and internally complain if they can't be found

lib/runtime.cpp
a-w50: Right, I need to rebase my refactor module branch, there only the manager reads the config.  This is really brittle here, because different module might have a different understanding about what the real config is

include/utils/config.hpp
hikinggrass: Sure, let's change them to const refs and keep them as strings to leave that option open for the future :)

lib/runtime.cpp
hikinggrass: That would be even better. If we could integrate a mechanism for providing a user config via the command line there I'll happily drop this PR in favor of a more robust solution.
Maybe we could even integrate a way of providing multiple user-configs (via the command line) that get applied on top of the main config in the specified order. This would make the config handling much more powerful.

PR #50, opened 2022-09-30
Fix controller only listening on localhost

hikinggrass:
This was making it unavailable from the outside


a-w50 (approved):
:100: 
maybe the commit could be named like
"Fix: let admin backend service listen on all devices"

PR #54, opened 2022-11-28
Refactoring config schema and command line parsing

a-w50:
- only prefix and config command line options are left for path/file specific settings
- everything else can be set in the config yaml
- config schema changes:
  - x-layout-* stuff moved to a separate section
  - plain module configuration is now under active_modules
  - config settings can be specified under settings
- implementation is still very inefficient, because the config gets loaded twice for each module, this will/could be reduced to only a single load
- compile time CMAKE_INSTALL_PREFIX gets included in the framework, so it will look for a good default

### DIFF
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -3,60 +3,210 @@
 
 #include <framework/runtime.hpp>
 
+#include <algorithm>
+
+#include <boost/program_options.hpp>
+
 namespace Everest {
-RuntimeSettings::RuntimeSettings(const po::variables_map& vm) : main_dir(vm["main_dir"].as<std::string>()) {
-    if (vm.count("schemas_dir")) {
-        schemas_dir = vm["schemas_dir"].as<std::string>();
+
+namespace po = boost::program_options;
+
+std::string parse_string_option(const po::variables_map& vm, const char* option) {
+    if (vm.count(option) == 0) {
+        return "";
+    }
+
+    return vm[option].as<std::string>();
+}
+
+static fs::path assert_dir(const std::string& path, const std::string& path_alias = "The") {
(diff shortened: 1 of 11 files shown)


lib/runtime.cpp
```diff
@@ -3,60 +3,211 @@
...
+        if (config_file_prefix.empty()) {
+            config_file_prefix = assert_dir(defaults::PREFIX, "Default prefix");
+        }
+
+        if (config_file_prefix.string() == "/usr") {
+            // we're going to look in /etc, which isn't prefixed by /usr
+            config_file_prefix = "/";
+        }
+
+        auto default_config_file = fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE / defaults::CONFIG_NAME;
```
hikinggrass: this is never used, only repeated a few lines onwardss

lib/runtime.cpp
```diff
@@ -3,60 +3,211 @@
...
+            const auto user_config_file =
+                config_file_prefix / defaults::SYSCONF_DIR / defaults::NAMESPACE / fmt::format("{}.yaml", config_);
+
+            const auto short_form_alias = fmt::format("User provided (by using short form: '{}')", config_);
+
+            config_file = assert_file(user_config_file, short_form_alias);
+        } else {
+            // default
+            config_file =
+                assert_file(config_file_prefix / defaults::SYSCONF_DIR / defaults::NAMESPACE / defaults::CONFIG_NAME,
```
hikinggrass, line 115: default config here?

hikinggrass (changes requested):
Looks good at first glance

hikinggrass (approved):

PR #56, opened 2022-12-01
Add python bindings for EVerest

corneliusclaussen:
- support for multiple requirements
- support for types
- support for module and implementation configs
- initial support for cross compiling
- publish variables with QOS2
- python modules use std::filesystem instead of boost::filesystem
- EVEREST_ENABLE_JS_SUPPORT and EVEREST_ENABLE_PY_SUPPORT options
  Set to "ON" by default these options make it possible to selectively enable and disable JavaScript and Python support in the framework
- Add everestpy to pylint extension-pkg-allow-list

Signed-off-by: Sebastian Lukas <sebastian.lukas@pionix.de>
Co-authored-by: Sebastian Lukas <sebastian.lukas@pionix.de>
Co-authored-by: aw <aw@pionix.de>


a-w50 (changes requested):
Could we move the QOS related stuff into a separate commit?
Its implementation by looking up the QOS key in *every* publish is not optimal.  Couldn't this be done only once?

hikinggrass:
> Could we move the QOS related stuff into a separate commit? Its implementation by looking up the QOS key in _every_ publish is not optimal. Couldn't this be done only once?

Done.
Concerning the implementation, as far as I can tell nobody is using the qos parameter at the moment. So we could drop the support for it from this PR altogether and re-introduce it with a general efficiency overhaul of the framework internals at a later date.

a-w50 (approved):
Ok, then lets drop the QOS commit.  I'll be fine with it

hikinggrass:
> Ok, then lets drop the QOS commit. I'll be fine with it

Lookup is removed, QOS2 is the default for variable publishing from now on. As was the case if you did not explicitly set a lower QOS in this branch for a long time

PR #63, opened 2023-01-04
Extended config settings

hikinggrass:


lib/runtime.cpp
```diff
@@ -208,7 +217,45 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string&
...
+    const auto settings_mqtt_port_it = settings.find("mqtt_port");
+    if (settings_mqtt_port_it != settings.end()) {
+        mqtt_port = (*settings_mqtt_port_it).get<std::string>();
+    } else {
+        mqtt_port = defaults::MQTT_PORT;
+    }
+
+    const auto settings_mqtt_everest_prefix_it = settings.find("mqtt_everest_prefix");
+    if (settings_mqtt_everest_prefix_it != settings.end()) {
+        mqtt_everest_prefix = (*settings_mqtt_everest_prefix_it).get<std::string>();
```
a-w50: why not settings_mqtt_everest_prefix_it->get ?

schemas/config.yaml
```diff
@@ -39,7 +39,19 @@ properties:
...
-      config_logging_file:
+      www_dir:
+        type: string
+      logging_config_file:
+        type: string
+      controller_port:
+        type: string
+      mqtt_server:
+        type: string
+      mqtt_port:
```
a-w50: why should port be of type string?

src/controller/controller.cpp
```diff
@@ -46,11 +46,12 @@ int main(int argc, char* argv[]) {
 
     RPC rpc(socket_fd, config);
     Server backend;
+    int controller_port = std::stoi(config_params.at("controller_port").get<std::string>());
```
a-w50: if port would be integer, we wouldn't need std::stoi

src/manager.cpp
```diff
@@ -327,7 +327,7 @@ static std::map<pid_t, std::string> start_modules(Config& config, MQTTAbstractio
         // FIXME (aw): implicitely adding ModuleReadyInfo and setting its ready member
         auto module_it = modules_ready.emplace(module_name, ModuleReadyInfo{false, nullptr}).first;
 
-        Handler module_ready_handler = [module_name, &mqtt_abstraction](nlohmann::json json) {
+        Handler module_ready_handler = [module_name, &mqtt_abstraction, &rs](nlohmann::json json) {
```
a-w50: be careful here, that rs doesn't get out of scope, furthermore, only rs.mqtt_everest_prefix is needed, this could be captured as well

src/manager.cpp
```diff
@@ -567,21 +569,10 @@ int boot(const po::variables_map& vm) {
...
-        mqtt_server_address = "localhost";
-    }
-
-    // NOLINTNEXTLINE(concurrency-mt-unsafe): not problematic that this function is not threadsafe here
-    const char* mqtt_server_port = std::getenv("MQTT_SERVER_PORT");
-    if (mqtt_server_port == nullptr) {
-        mqtt_server_port = "1883";
-    }
-
-    MQTTAbstraction& mqtt_abstraction = MQTTAbstraction::get_instance(mqtt_server_address, mqtt_server_port);
```
a-w50, line 595: I'm not sure, if we should leave the possibility to use environment variables for configuring the mqtt server, otherwise one would need to touch the config every time. But on the other hand, if we will need it again, we can add it back

include/framework/everest.hpp
```diff
@@ -137,30 +140,17 @@ class Everest {
 
     ///
     /// \returns the instance of the Everest singleton taking a \p module_id, the \p config, a \p mqtt_server_address
-    /// and \p mqtt_server_port as parameters. If validation of data with the known json schemas is needed this can be
-    /// activated by setting \p validate_data_with_schema to true
+    /// , \p mqtt_server_port , \p mqtt_everest_prefix and \p mqtt_external_prefix as parameters. If validation of data
+    /// with the known json schemas is needed this can be activated by setting \p validate_data_with_schema to true
     static Everest& get_instance(std::string module_id, Config config, bool validate_data_with_schema,
```
a-w50, line 145: this got quite large, maybe all the parameters should go into a meaningful struct?, actually most of the parameters could be part of the Config parameter

everestpy/everestpy.cpp
```diff
@@ -231,8 +220,7 @@ int initialize(const std::string& prefix, const std::string& config_file, const
         }
 
         if (!everest.connect()) {
-            EVLOG_critical << fmt::format("Cannot connect to MQTT broker at {}:{}", mqtt_server_address,
-                                          mqtt_server_port);
+            EVLOG_critical << fmt::format("Cannot connect to MQTT broker at {}:{}", rs.mqtt_server, rs.mqtt_port);
```
a-w50: concerning naming - choosing mqtt_server_address was my fault, because an address includes the port, if we use mqtt_port, than I would prefer mqtt_host instead of mqtt_server

include/framework/runtime.hpp
```diff
@@ -70,6 +70,12 @@ inline constexpr auto LOGGING_CONFIG_NAME = "default_logging.cfg";
 
 inline constexpr auto WWW_DIR = "www";
 
+inline constexpr auto CONTROLLER_PORT = "8849";
+inline constexpr auto MQTT_SERVER = "localhost";
+inline constexpr auto MQTT_PORT = "1883";
+inline constexpr auto MQTT_EVEREST_PREFIX = "everest/";
```
a-w50: i would prefer to skip the trailing slash here

a-w50: maybe even call it MQTT_EVEREST_ROOT_TOPIC

lib/runtime.cpp
hikinggrass: why not indeed

src/manager.cpp
hikinggrass: I've re-added this to the runtime settings which should reduce the code duplication whilst retaining this feature

everestpy/everestpy.cpp
hikinggrass: I've no settled on mqtt_broker_host and mqtt_broker_port to make it really clear what is meant by these settings

include/framework/runtime.hpp
hikinggrass: Trailing slash is gone here, runtime settings will now check if the prefix ends with a trailing slash, and if not add one.

I've kept the name for now, but that might just be personal preference

include/framework/everest.hpp
hikinggrass: Absolutely, I would like to do some general restructuring here at some point, there's quite a lot of code duplication in various parts  concerning these parameters. Especially in the JavaScript and Python binding code

lib/everest.cpp
```diff
@@ -20,10 +20,10 @@ namespace Everest {
 const auto remote_cmd_res_timeout_seconds = 300;
 
 Everest::Everest(std::string module_id, Config config, bool validate_data_with_schema,
-                 const std::string& mqtt_server_address, const std::string& mqtt_server_port,
-                 const std::string& mqtt_everest_prefix, const std::string& mqtt_external_prefix) :
-    mqtt_abstraction(MQTTAbstraction::get_instance(mqtt_server_address, mqtt_server_port, mqtt_everest_prefix,
-                                                   mqtt_external_prefix)),
+                 const std::string& mqtt_server_address, int mqtt_server_port, const std::string& mqtt_everest_prefix,
```
a-w50, line 23: couldn't we also use mqtt_broker_host and mqtt_broker_port here?

hikinggrass: absolutely, I'll change that in an upcoming commit

a-w50:
poke @hikinggrass 
what's the current status with this PR?

lib/everest.cpp
hikinggrass: I would leave this for now and tackle this in an upcoming PR with more cleanup

hikinggrass:
> poke @hikinggrass what's the current status with this PR?

I would like to merge it as-is and do the restructuring in a new PR

a-w50 (approved):
compiles fine :+1:

PR #65, opened 2023-01-12
Use find_package for pybind11 when not using EDM

hikinggrass:


a-w50 (approved):
:+1:

PR #67, opened 2023-01-17
Remove special cross compilation handling of everestpy and update pybind11

hikinggrass:
This fixes an issue with pybind11 not finding the appropriate python environment during cross compilation


CMakeLists.txt
```diff
@@ -29,6 +29,17 @@ find_package(Boost
...
+set(EVEREST_CROSS_PYTHON_INCLUDE_DIRS "${CMAKE_SYSROOT}/usr/include/python3.9" CACHE STRING "Path to python include directory")
+set(EVEREST_CROSS_PYTHON_LIBRARIES "${CMAKE_SYSROOT}/usr/lib/arm-linux-gnueabihf/libpython3.9.so" CACHE STRING "Path to python library")
+
+if (CMAKE_CROSSCOMPILING)
+    set(PYTHON_INCLUDE_DIRS ${EVEREST_CROSS_PYTHON_INCLUDE_DIRS} CACHE INTERNAL "")
+    set(PYTHON_LIBRARIES ${EVEREST_CROSS_PYTHON_LIBRARIES} CACHE INTERNAL "")
+    set(PYTHON_MODULE_EXTENSION ".so" CACHE INTERNAL "")
+
+    set(PYTHONLIBS_FOUND TRUE CACHE INTERNAL "")
+endif()
```
a-w50: are these EVEREST_CROSS variables used anywhere else? If not, I would drop them and set the PYTHON_* variables directly, instead of setting them with the EVEREST_CROSS variables.  Furthermore, does yocto build with this hardcoded setup? I doubt it, therefore I would put it after `if(NOT DISABLE_EDM)`, but before `evc_setup_edm()`.  There is also https://cmake.org/cmake/help/latest/module/FindPython3.html which could be possibly used to find the correct files right away

a-w50 (changes requested):

CMakeLists.txt
hikinggrass: At the moment these EVEREST_CROSS_* variables are used in the kirkstone-next branch of meta-everest. I would also prefer to either set the PYTHON_* variables directly or investigate the use of the FindPython3 module to do this automatically, this was just the least intrusive and quickest way of fixing cross compiling earlier

a-w50 (approved):

PR #68, opened 2023-01-18
Log a message indicating all modules excluding standalone ones are ready

hikinggrass:

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -316,8 +316,6 @@ static std::map<pid_t, std::string> start_modules(Config& config, MQTTAbstractio
     auto main_config = config.get_main_config();
     modules_to_spawn.reserve(main_config.size());
 
-    bool modules_spawned = false;
-
     for (const auto& module : main_config.items()) {
         std::string module_name = module.key();
         if (std::any_of(ignored_modules.begin(), ignored_modules.end(),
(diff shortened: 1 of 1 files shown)


a-w50 (approved):
Looks good.  In case multiple standalone modules are loaded, the message will be repeated right?

hikinggrass:
> Looks good. In case multiple standalone modules are loaded, the message will be repeated right?

Yes, but I'm thinking of changing this to be able to log more detailed output when a standalone module registers itself with the manager

a-w50 (changes requested):
I think, the problem with modules_spawned here is, that the variable gets out of scope, when the outer function returns, and it's passed by reference to the lambda. This would lead to undefined behaviour. The variable could be made static, to solve the problem but it needs to be assured that the outer function isn't called before all the module_ready_handlers are done (thats actually the content of the FIXME concerning the race conditions)

src/manager.cpp
```diff
@@ -337,11 +337,22 @@ static std::map<pid_t, std::string> start_modules(Config& config, MQTTAbstractio
...
+            }
             if (std::all_of(modules_ready.begin(), modules_ready.end(),
                             [](const auto& element) { return element.second.ready; })) {
                 EVLOG_info << fmt::format(TERMINAL_STYLE_OK,
                                           ">>> All modules are initialized. EVerest up and running <<<");
                 mqtt_abstraction.publish("everest/ready", nlohmann::json(true));
+            } else if (!standalone_modules.empty()) {
+                auto modules_spawned = std::count_if(modules_ready.begin(), modules_ready.end(),
+                                                     [](const auto& element) { return element.second.ready; });
+                if (modules_spawned == modules_ready.size() - standalone_modules.size()) {
```
a-w50, line 354: The code is not completely correct, because it could also happen, that a standalone module starts before one of the automatically ones and then the message would be incorrect, but I doubt this will ever happen or have any impact.

a-w50 (approved):

src/manager.cpp
hikinggrass: The manager probably would not even realize that such a standalone module was started early because it needs to have the module ready topic subscribed before a standalone module can start. But yes, there is a small race condition here that most likely will not matter in the usecase for this particular functionality (starting a module by-hand for debugging or testing purposes)

a-w50: :+1:

PR #73, opened 2023-02-22
Add telemetry mechanism

hikinggrass:


corneliusclaussen (approved):

lib/everest.cpp
```diff
@@ -15,28 +15,33 @@
 #include <date/tz.h>
 #include <framework/everest.hpp>
 #include <utils/conversions.hpp>
+#include <utils/date.hpp>
 
 namespace Everest {
 const auto remote_cmd_res_timeout_seconds = 300;
+const std::array<std::string, 3> TELEMETRY_RESERVED_KEYS = {{"connector_id"}};
```
a-w50, line 22: type and timestamp should be also here, right?

hikinggrass: I actually wanted to be able to sometimes overwrite at least the timestamp, for example when the data is slightly older (for example power meter data that can already come with a timestamp). But you're right, I must have overlooked re-adding the type to this list

PR #74, opened 2023-02-23
Increase version to 0.4

hikinggrass:
There have been a few changes to the API of everestjs and everestpy as well as the addition of new settings variables. I think this requires a new version number, so we can properly reference these changes in everest-core


hikinggrass:
(this is only needed if we don't merge #73 before the 2023.2 release)

corneliusclaussen:
We can merge #73 before the release

PR #75, opened 2023-03-03
Performance enhancements

hikinggrass:


lib/message_queue.cpp
```diff
@@ -58,14 +58,13 @@ MessageQueue::~MessageQueue() {
...
         while (true) {
-            std::shared_ptr<json> message;
             std::unique_lock<std::mutex> lock(this->handler_ctrl_mutex);
             this->cv.wait(lock, [this]() { return !this->message_queue.empty() || this->running == false; });
             if (!this->running) {
                 return;
             }
 
-            message = this->message_queue.front();
+            std::shared_ptr<json> message = this->message_queue.front();
```
a-w50:
```
auto message = std::move(this->message_queue.front());
this->message_queue.pop();
```
would be even more performant

a-w50 (approved):
Nice cleanups

lib/message_queue.cpp
hikinggrass: Good idea!

PR #76, opened 2023-04-03
Fix libwebsockets build with newer gcc

a-w50:


corneliusclaussen (approved):
works with
gcc (SUSE Linux) 13.0.1 20230314 (experimental) [revision 42630fadbe248717859d61c0244c821c32b4e52c]

PR #79, opened 2023-04-28
Enhanced ModuleInfos path properties

a-w50:
- ModuleInfo now carries a path property which has the sub-properties:
- etc -> determined etc folder
- libexec -> determined {libexec}/everest/modules/ModuleName folder
- share -> determined {share}/everest/modules/ModuleName folder


SebaLukas (approved):
Really good 👍

hikinggrass (approved):

PR #80, opened 2023-05-03
Added optional compile warnings + some fixes

a-w50:
- add possibility to specify compiler warnings
- in order to switch on compile warnings for source files, use cmake ... -DCOMPILER_WARNING_OPTIONS="-Wall;-Wextra;-Wpedantic" this will add these flags to all source files of this repository targets
- a few of the resulting warnings have been resolved, some are still there and need to be fixed

### DIFF
--- a/lib/serial.cpp
+++ b/lib/serial.cpp
@@ -117,6 +117,7 @@ uint32_t Serial::crc32(uint8_t* buf, int len) {
 }
 
 void Serial::handlePacket(uint8_t* buf, int len) {
+    // FIXME (aw): why is there no implementation here?
 }
 
 void Serial::cobsDecode(uint8_t* buf, int len) {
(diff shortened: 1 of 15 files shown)


src/controller/server.cpp
```diff
@@ -119,13 +119,32 @@ const lws_protocol_vhost_options Server::Impl::pvo = {nullptr, &pvo_opt, "everes
 
 const lws_protocol_vhost_options Server::Impl::pvo_mime = {nullptr, nullptr, ".mp4", "application/x-mp4"};
 
+// lws_http_mount Server::Impl::mount = {
```
hikinggrass: why keep this around?

lib/serial.cpp
```diff
@@ -117,6 +117,7 @@ uint32_t Serial::crc32(uint8_t* buf, int len) {
 }
 
 void Serial::handlePacket(uint8_t* buf, int len) {
+    // FIXME (aw): why is there no implementation here?
```
hikinggrass, line 120: we should make this virtual, probably does not make much sense to have a default implementation

src/controller/server.cpp
a-w50: sorry, forgot to delete that

lib/serial.cpp
a-w50: yes, I guess this should be done in a separate refactor for the serial lib.  Currently Yeti and uMWC for example still don't use it and have their own `openSerial` implementation

hikinggrass (approved):

PR #83, opened 2023-05-14
Added status notification support via named pipe

a-w50:
- for better control, especially when testing, the manager now supports the option "status-fifo", which shall be a path to a named pipe (created via mkfifo)
- this file needs to be opened for reading (in order to listen on status updates of the manager), when the manager gets started, otherwise it will terminate assuming that the caller did something unintended
- on status updates like "all modules started" or similar, the manager will print simple strings on this pipe
- the manager does not read from this pipe
- if the caller stops reading from the pipe, the manager continues running but stops writing to the pipe when it becomes aware of that there is no reader anymore on the pipe

### DIFF
--- a/include/utils/status_fifo.hpp
+++ b/include/utils/status_fifo.hpp
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+#ifndef STATUS_FIFO_HPP
+#define STATUS_FIFO_HPP
+
+#include <string>
+
+namespace Everest {
+
+class StatusFifo {
+public:
+    // defined messages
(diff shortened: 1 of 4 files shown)


include/utils/status_fifo.hpp
```diff
@@ -0,0 +1,36 @@
...
+    StatusFifo(StatusFifo const&) = delete;
+    StatusFifo& operator=(StatusFifo const&) = delete;
+    // NOTE (aw): the move constructor could be implementented, but we don't need it for now
+    StatusFifo(StatusFifo&&) = delete;
+    StatusFifo& operator=(StatusFifo&&) = delete;
+    ~StatusFifo();
+
+private:
+    StatusFifo() = default;
+    StatusFifo(int fd_) : fd(fd_), disabled(false), opened(true){};
```
hikinggrass: maybe fix this codacy issue, otherwise it looks good!

hikinggrass (approved):

PR #92, opened 2023-07-13
Feature/error handling

andistorm:


andistorm:
depends on https://github.com/EVerest/everest-utils/pull/59

cmake/convert-refs.cmake
```diff
@@ -0,0 +1,85 @@
+include(ev-cli)
+require_ev_cli_version("0.0.20")
```
hikinggrass: This should require 0.0.21

dependencies.yaml
```diff
@@ -73,3 +73,6 @@ pybind11:
 pybind11_json:
   git: https://github.com/pybind/pybind11_json.git
   git_tag: 0.2.13
+everest-utils:
+  git: https://github.com/EVerest/everest-utils.git
+  git_tag: v0.1.4
```
hikinggrass: This should at least reference the most recent commit on the related feature branch, unfortunately we track the version of everest-testing there, not ev-cli

cmake/convert-refs.cmake
```diff
@@ -0,0 +1,85 @@
...
+        set (CONVERT_REFS_INSTALL_DIR ${ARGV2})
+    else()
+        message(FATAL_ERROR "${CMAKE_CURRENT_FUNCTION}() expects exactly three arguments")
+    endif()
+
+    get_filename_component(CONVERT_REFS_INPUT_FILE_NAME ${CONVERT_REFS_INPUT_FILE} NAME)
+
+    get_filename_component(CONVERT_REFS_INPUT_FILE_NAME_WITHOUT_EXT ${CONVERT_REFS_INPUT_FILE} NAME_WE)
+    get_filename_component(CONVERT_REFS_INPUT_FILE_PARENT ${CONVERT_REFS_INPUT_FILE} DIRECTORY)
+    cmake_path(GET CONVERT_REFS_INPUT_FILE_PARENT STEM CONVERT_REFS_INPUT_FILE_PARENT_NAME)
```
hikinggrass: This only works with cmake >=3.20 and we target 3.14 so I think we should use get_filename_component here as well, or is this slightly different in behavior?
```cmake
get_filename_component(CONVERT_REFS_INPUT_FILE_PARENT_NAME ${CONVERT_REFS_INPUT_FILE_PARENT} NAME_WE)
```

lib/config.cpp
hikinggrass, line 1: Should we load all files in errors/ here as well and validate them against the schema? (only in the manager for performance reasons)

lib/mqtt_abstraction_impl.cpp
```diff
@@ -218,14 +218,7 @@ void MQTTAbstractionImpl::on_mqtt_message(std::shared_ptr<Message> message) {
         std::vector<Handler> local_handlers;
         for (auto& [handler_topic, handler] : this->message_handlers) {
             bool topic_matches = false;
-            if (is_everest_topic) {
```
hikinggrass: Not having this shortcut would be a noticeable performance regression

src/manager.cpp
```diff
@@ -381,6 +385,9 @@ static std::map<pid_t, std::string> start_modules(Config& config, MQTTAbstractio
 
         mqtt_abstraction.register_handler(topic, module_it->second.token, QOS::QOS2);
 
+        std::string error_topic = fmt::format("{}/+/error/#", module_name);
```
hikinggrass: Do we have to use wildcards here?

hikinggrass (changes requested):

src/manager.cpp
andistorm: I changed it to adding each topic one by one. A restructuring of the topic structure in the future could increase performance here probably

lib/config.cpp
```diff
@@ -518,14 +563,73 @@ json Config::resolve_interface(const std::string& intf_name) {
...
+    return interface_json;
+}
+
 json Config::load_interface_file(const std::string& intf_name) {
     BOOST_LOG_FUNCTION();
     fs::path intf_path = this->rs->interfaces_dir / (intf_name + ".yaml");
     try {
         EVLOG_debug << fmt::format("Loading interface file at: {}", fs::canonical(intf_path).string());
 
         json interface_json = load_yaml(intf_path);
```
andistorm:
```suggestion
        json interface_json = load_yaml(intf_path);

```

lib/everest.cpp
```diff
@@ -24,7 +26,8 @@ const std::array<std::string, 3> TELEMETRY_RESERVED_KEYS = {{"connector_id"}};
 
 Everest::Everest(std::string module_id_, const Config& config_, bool validate_data_with_schema,
                  const std::string& mqtt_server_address, int mqtt_server_port, const std::string& mqtt_everest_prefix,
-                 const std::string& mqtt_external_prefix, const std::string& telemetry_prefix, bool telemetry_enabled) :
+                 const std::string& mqtt_external_prefix, const std::string& telemetry_prefix, bool telemetry_enabled,
+                 const fs::path& errors_dir) :
```
hikinggrass: since the errors_dir is already passed to the Config, couldn't we do the generation of the ErrorTypeMap there? So we would not have to change the constructor here

lib/everest.cpp
```diff
@@ -370,6 +375,230 @@ void Everest::subscribe_var(const Requirement& req, const std::string& var_name,
...
+        std::make_shared<TypedHandler>(error_type, HandlerType::SubscribeError, std::make_shared<Handler>(handler));
+    this->mqtt_abstraction.register_handler(error_cleared_topic, token, QOS::QOS2);
+}
+
+std::string Everest::raise_error(const std::string& impl_id, const std::string& error_type, const std::string& message,
+                                 const std::string& severity) {
+    BOOST_LOG_FUNCTION();
+
+    std::string description = this->error_map.get_description(error_type);
+    bool persistent = false;
```
hikinggrass: codacy

lib/error_database.cpp
```diff
@@ -0,0 +1,82 @@
...
+namespace error {
+
+ErrorHandle ErrorDatabase::add_error(std::shared_ptr<Error> error) {
+    BOOST_LOG_FUNCTION();
+
+    ErrorHandle handle(error->uuid);
+    if (this->has_error(handle)) {
+        EVTHROW(EverestErrorAlreadyExistsError(handle));
+    }
+    this->errors[handle] = error;
```
hikinggrass, line 24: Access to this map should be protected by a muted so we don't add/remove items whilst iterating over it

hikinggrass (approved):

PR #99, opened 2023-09-15
Rust integration (experimental)

SirVer:
This is using [cxx.rs](https://cxx.rs) to wrap the framework library and calling it from Rust. This code is currently only supporting providing an interface to be implemented, i.e. no variables publish or receiving and no calling of other interfaces. Those features are straightforward, quick and easy to implement, but for now this is probably enough to iron out the integration questions.

See the everestrs/README.md for more details on the implementation.

Co-authored-by: Dima Dorezyuk <ddo@qwello.eu>


everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -0,0 +1,83 @@
...
+      rs.interfaces_dir.string(), rs.types_dir.string(), rs.mqtt_everest_prefix,
+      rs.mqtt_external_prefix);
+}
+
+JsonBlob json2blob(const json& j) {
+  // I did not find a way to not copy the data at least once here.
+  const std::string dumped = j.dump();
+  rust::Vec<uint8_t> vec;
+  vec.reserve(dumped.size());
+  std::copy(dumped.begin(), dumped.end(), std::back_inserter(vec));
```
dorezyuk: not really what you want but std::move should make one copy less

everestrs/everestrs/build.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Takes a path to a library like `libframework.so` and returns the name for the linker, aka
+/// `framework`
+fn libname_from_path(p: &Path) -> String {
+    let mut name = p
+        .file_stem()
+        .and_then(|os_str| os_str.to_str())
+        .unwrap_or("")
+        .to_string();
+    name.drain(..3); // remove 'lib' prefix.
```
dorezyuk: https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix should be better and easier to understand

everestrs/everestrs/build.rs
```diff
@@ -0,0 +1,89 @@
...
+            framework,
+        })
+    } else {
+        None
+    }
+}
+
+/// Returns the Libraries path if this is an EVerest workspace where make install was run in
+/// everest-core/build or None if not.
+fn find_libs_in_qwello_core_build_dist(root: &Path) -> Option<Libraries> {
```
dorezyuk: not sure why its called qwello

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    fn on_ready(&mut self) {}
+}
+
+pub struct Runtime<T: GenericModule> {
+    // We are handing out pointers to `module_impl` to `cpp_module` for callbacks. The pointers must
+    // must stay valid for as long as `cpp_module` is alive. Hence `module_impl` must never move in
+    // memory. Rust can model this through the Pin concept which upholds this guarantee. We use a
+    // Box to put the object on the heap.
+    module_impl: Pin<Box<T>>,
```
dorezyuk: Rather a niptick but may be painful at some point in time - rust drops the fields in declaration order so you want to make sure that the cpp_module is dropped before module_impl

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -0,0 +1,83 @@
...
+JsonBlob Module::initialize() {
+  handle_->connect();
+  handle_->spawn_main_loop_thread();
+
+  const std::string& module_name =
+      config_->get_main_config().at(module_id_).at("module");
+  return json2blob(config_->get_manifests().at(module_name));
+}
+
+void Module::signal_ready(const char* obj,
```
dorezyuk: I'm wondering can't this already take a function pointer? so you pass here a lambda from rust code? This would make the code clearer since we can get rid of the obscure *char pointer. Otherwise it should be probably a void pointer to indicate that we're passing just some address

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Handler for the command `name` on `implementation_id` with the given `parameters`. The return value
+    /// will be returned as the result of the call.
+    fn handle_command(
+        &mut self,
+        implementation_id: &str,
+        name: &str,
+        parameters: HashMap<String, serde_json::Value>,
+    ) -> Result<serde_json::Value>;
+
+    fn on_ready(&mut self) {}
```
dorezyuk: Not sure but I don't expect that the rust compiler can check if someone is holding multiple mutable refs to this struct - since these are called from within the c++ code. I would suggest to change the signature to `&self` such that if users are changing data in the callback they are forced to wrap it in a mutex

everestrs/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+endif()
+
+emit_cxxrs_header()
+emit_cxxrs_for_module(everestrs)
+
+add_library(everestrs_sys STATIC
+  ${CMAKE_CURRENT_BINARY_DIR}/cxxbridge/everestrs/lib.rs.cc
+  everestrs_sys/everestrs_sys.cpp
+)
+
```
SebaLukas, line 21: Without this option the linker gives an error when building the module:
`set_property(TARGET everestrs_sys PROPERTY POSITION_INDEPENDENT_CODE ON)`

everestrs/everestrs_sys/everestrs_sys.cpp
SirVer: Not sure where the std::move would need to go here... std::back_inserter is taking a reference already and the move into the `JsonBlob` is elided by the compiler anyways, see https://godbolt.org/z/986cWfMqs.

everestrs/everestrs/build.rs
SirVer: good call. Done.

everestrs/everestrs/src/lib.rs
SirVer: Very good point. I do not think it is easy to enforce without complicating the code, I have hence only added a comment explaining it and reordered the fields.

everestrs/everestrs_sys/everestrs_sys.cpp
SirVer: 1. As discussed in person, it is not possible to pass a pointer to a lambda to C, since the calling conventions of Rust are different than the ones of C. The only way forward is to pass a function pointer with C conventions that gets passed a pointer to a Rust object, which can be casted in Rust into the right type and then used. This is widely unsafe of course. 

2. unfortunately, cxx.rs only exposes `char*`, not any other pointer type. I added a comment. 🤷

everestrs/everestrs/src/lib.rs
SirVer: Yes. For now I think the code is actually correct, since the Everest runtime only ever calls one callback at the same time from the same thread, however @dorezyuk and me discussed in person and realized that if calling into the module implementation from another thread in Rust land we might run into race conditions. We hence feel that the correct design is to rely on internal mutability (i.e. taking a `&self` instead of a `&mut self` in all modules and have the user wrap their data in a Mutex). This is also how other projects like Tonic handle this problem.

I changed the code accordingly.

everestrs/everestrs/build.rs
SirVer: brainfart..... fixed.

everestrs/everestrs/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "everestrs"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+argh = "0.1.10"
+cxx = { version = "1.0.107", features = ["c++20"] }
```
SebaLukas: c++17 instead of c++20, because Everest uses c++17

SirVer:
All comments addressed. The cpp linter is wrong btw, `module_id_` is used: https://github.com/sirver/everest-framework/blob/hrapp/initial_rust_support/everestrs/everestrs_sys/everestrs_sys.cpp#L58

cmake/cxxrs.cmake
```diff
@@ -0,0 +1,19 @@
...
+  set(output_dir "${CMAKE_CURRENT_BINARY_DIR}/cxxbridge/${module_name}/")
+  file(MAKE_DIRECTORY ${output_dir})
+
+  add_custom_command(
+      OUTPUT ${output_dir}lib.rs.h ${output_dir}lib.rs.cc
+      COMMAND ${CXXBRIDGE} ${CMAKE_CURRENT_SOURCE_DIR}/${module_name}/src/lib.rs --header -o ${output_dir}lib.rs.h
+      COMMAND ${CXXBRIDGE} ${CMAKE_CURRENT_SOURCE_DIR}/${module_name}/src/lib.rs -o ${output_dir}lib.rs.cc
+      DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${module_name}/src/lib.rs
+      COMMENT "Generating cxx for ${module_name}"
+      VERBATIM
```
a-w50, line 17: - rather than calling the cxxbridge util here directly, I would call `cxx_build::bridge` from within the `build.rs`
- I think, it should be the concern of this `cxxrs.cmake` script to figure out the path of `CXXBRIDGE` and not if its including script

everestrs/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+target_include_directories(everestrs_sys PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  ${CMAKE_CURRENT_BINARY_DIR}/cxxbridge
+)
+
+# This is a requirement that linking works on systems enforcing PIE.
+set_property(TARGET everestrs_sys PROPERTY POSITION_INDEPENDENT_CODE ON)
```
a-w50, line 28: Not sure, if PIC here is really necessary, as the resulting binary executable shouldn't rely on that and this static library isn't linked into a shared library.  Probably some linking misconfiguration, if this is really needed.

everestrs/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+target_link_libraries(everestrs_sys
+    PRIVATE
+    everest::framework
+    everest::log
+)
+
+install(TARGETS everestrs_sys LIBRARY)
```
a-w50, line 35: I wouldn't binary ship this glue code library, as it depends on auto-generated stuff (`rust/cxx.h'), which will only be generated, when its *using* crate is build.

everestrs/Cargo.lock
```diff
@@ -0,0 +1,196 @@
...
+
+[[package]]
+name = "argh"
+version = "0.1.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7af5ba06967ff7214ce4c7419c7d185be7ecd6cc4965a8f6e1d8ce0398aad219"
+dependencies = [
+ "argh_derive",
+ "argh_shared",
+]
```
a-w50, line 13: not sure about rust best practice here, but should this lock file be checked in here?

everestrs/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
+resolver = "2"
+members = [
+   "everestrs",
+]
```
a-w50, line 5: the workspace shouldn't be necessary for this single everestrs crate, right?

everestrs/everestrs/build.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Takes a path to a library like `libframework.so` and returns the name for the linker, aka
+/// `framework`
+fn libname_from_path(p: &Path) -> String {
+    p.file_stem()
+        .and_then(|os_str| os_str.to_str())
+        .expect("'p' must be valid UTF-8 and have a .so extension.")
+        .strip_prefix("lib")
+        .expect("'p' should start with `lib`")
+        .to_string()
+}
```
a-w50, line 63: This functionality is quite brittle due to hardcoded paths, also depending on cmake internals.  As shown, I think it would be easier to let cmake pass its knowledge about the library paths and names to cargo or let cmake create a file which contains this information.  Of course this would mean, cmake needs to be run in order to let cargo do its work, but probably this has to be done anyway, because libframework needs to be build beforehand.

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+#[cxx::bridge]
+mod ffi {
+    struct CommandMeta {
+        implementation_id: String,
+        name: String,
+        // cxx.rs does not expose a *const c_void type, so we use *const c_char everywhere.
+        // Wherever this shows it means "pointer to some random blob of memory that only Rust
+        // understand what it means and is of no concern to C".
+        obj: *const c_char,
```
a-w50: Not sure, if this is the only way, but it doesn't look right, to pass this "void" pointer around (see the comment further below).

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    struct JsonBlob {
```
a-w50, line 33: Not sure about this, but wouldn't a type alias enough here?

a-w50: Ah, then you won't be able to implement functions on this type.  Using free functions would still be an option ...

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    fn deserialize<T: DeserializeOwned>(self) -> T {
+        // TODO(hrapp): Error handling
+        serde_json::from_slice(self.as_bytes()).unwrap()
+    }
+
+    fn from_vec(data: Vec<u8>) -> Self {
+        Self { data }
+    }
+}
```
a-w50, line 73: If using a type alias, you can drop the `as_bytes` and `from_vec` functions.

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        &self,
+        implementation_id: &str,
+        name: &str,
+        parameters: HashMap<String, serde_json::Value>,
+    ) -> Result<serde_json::Value>;
+
+    fn on_ready(&self) {}
+}
+
+pub struct Runtime<T: GenericModule> {
```
a-w50: This is like the runtime handle, which keeps the rust module alive and tracks its link to the cxx side "module".
Concerning the naming: we used `Runtime` usually for the runtime setup, that is, where is everest located (prefix) and the logging config I think.  This runtime can then be used to spawn a new module (or in everestpy even multiple modules).  I personally like this flexibility to create/implement various modules inside one unit for running tests and things like that.
Concerning the link between both, why not having the rust module struct having it's one member variable, linking to the cxx module?

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    // 2. For the same reason, `module_impl` should outlive `cpp_module`, hence should be dropped
+    //    after it. Rust drops fields in declaration order, hence `cpp_module` should come before
+    //    `module_impl` in this struct.
+    cpp_module: cxx::UniquePtr<ffi::Module>,
+    module_impl: Pin<Box<T>>,
+}
+
+impl<T: GenericModule> Runtime<T> {
+    fn ptr_to_module_impl(&self) -> *const c_char {
+        &*self.module_impl.as_ref() as *const T as *const c_char
```
a-w50: This definitely looks like code smell.  I see the point, that you want to access the module handle inside the lambda functions via the command meta in order to do thinks, but why don't you access the module via capture in the lambda?  Furthermore, for all callbacks, you need to add this pointer to the arguments.

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        let manifest_json = cpp_module.as_mut().unwrap().initialize();
+        let manifest: schema::Manifest = manifest_json.deserialize();
+        let module = Self {
+            cpp_module,
+            module_impl: Box::pin(module_impl),
+        };
+
+        // Implement all commands for all of our implementations, dispatch everything to the
+        // GenericModule.
+        for (implementation_id, provides) in manifest.provides {
```
a-w50: I would prefer `for (implementation_id, implementation) in manifest.provides {`

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            let interface_s = module.cpp_module.get_interface(&provides.interface);
```
a-w50: This kind of information encapsulation is a bit problematic.  Unfortunately, in everest-framework, we still keep the structures about which module connects to which and how it is configured as json, instead of proper typed models.
The language bindings, ideally shouldn't care about the full json schema of the manifest / interfaces etc.  They really only should care about the necessary stuff, like their connections and and config options etc.
Because otherwise, like in this case, the binding depends on implementation details (like how the schema looks like etc).

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            let interface: schema::Interface = interface_s.deserialize();
+            for (name, _) in interface.cmds {
+                let meta = ffi::CommandMeta {
+                    implementation_id: implementation_id.clone(),
+                    name,
+                    obj: module.ptr_to_module_impl(),
```
a-w50: Not sure with Rust, but why can't you capture `module_impl` inside the lambda?  I don't see why this shouldn't be possible?

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    //    move in memory. Rust can model this through the Pin concept which upholds this guarantee.
+    //    We use a Box to put the object on the heap.
+    // 2. For the same reason, `module_impl` should outlive `cpp_module`, hence should be dropped
+    //    after it. Rust drops fields in declaration order, hence `cpp_module` should come before
+    //    `module_impl` in this struct.
+    cpp_module: cxx::UniquePtr<ffi::Module>,
+    module_impl: Pin<Box<T>>,
+}
+
+impl<T: GenericModule> Runtime<T> {
```
a-w50: Is it really necessary to have the `GenericModule` a generic?  Of course there will be specialized implementations, but due to the autogenerated module part later, there is again a level of indirection which can capture this generalization.

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+                let meta = ffi::CommandMeta {
+                    implementation_id: implementation_id.clone(),
+                    name,
+                    obj: module.ptr_to_module_impl(),
+                };
+                unsafe {
+                    module.cpp_module.as_ref().unwrap().provide_command(
+                        meta,
+                        |meta: &ffi::CommandMeta, params: ffi::JsonBlob| {
+                            let module_impl = &mut *(meta.obj as *mut T);
```
a-w50: as mentioned above, why not using capturing here?

everestrs/everestrs/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        unsafe {
+            module
+                .cpp_module
+                .as_ref()
+                .unwrap()
+                .signal_ready(module.ptr_to_module_impl(), |obj| {
+                    let module_impl = &mut *(obj as *mut T);
+                    module_impl.on_ready();
+                });
+        }
```
a-w50: This is not how we intended to use the `signal ready`.  Here it gets called automatically.  The idea was the following:
- the module *says* hello to everest and tells it's module instance id
- in return it should get its config and its fullfillments (i.e. the connections, which satisfy its requirements)
- based on this information, it can implement / register its handlers (in python and js, this is dynamic), and it could also be done dynamic in cxx and rust (the current rust implementation doesn't allow for this I think, because traits are used for the interface, of course this could be dispatched again ...)
- after registering the handlers and some other user stuff (for example for RiseV2G we needed to wait for the java process getting successful spawned) the *init done* or *i am ready* should be called on everest.  So calling the ready should be done by the user and not automatically like here, because otherwise there is no way for the implemented handlers to depend timewise on other things (of course they could block later on, but thats another strategy we hadn't in mind so far)
- finally, if all module sent their ready, their `on_ready` should get called

everestrs/everestrs/src/schema/interface.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub vars: BTreeMap<String, Variable>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct Command {
+    pub description: String,
+    #[serde(default)]
+    pub arguments: BTreeMap<String, Variable>,
+    pub result: Option<Variable>,
+}
```
a-w50, line 19: as mentioned above, these are "implementation details" from my point of view, and the binding should care about that. But currently libframework doesn't have it's own model structures for passing only the necessary information ..

a-w50: See here https://github.com/EVerest/everest-framework/blob/wip/refactor/module_interface/include/everest/utils/schema.hpp
This is a refactor I once made which strongly types all the necessary information in c++ and only exposes the necessary things to the binding.

cmake/cxxrs.cmake
SirVer: 1. The first one is a chicken and egg though: You would need to run cargo build to get the cpp code, which you need to run cmake to build the static C library, which cargo needs to link against. I.e. you would need to run cargo for only the build.rs file (which I do not think is possible), then run cmake, then cargo again. I think doing it here is the only possibility.
2. I do not understand this comment. Can you expand?

everestrs/CMakeLists.txt
SirVer: I did not need it on my system, however it was needed for @SebaLukas on his system. I do not understand what is going on under the hood to comment if it is necessary or not - however I have worked on systems were PIE was enforced for all code. Not sure what to do here to resolve this.

everestrs/CMakeLists.txt
SirVer: I do understand that you do not like the installing piece, however the comment is not actionable to me, I do not know what to do instead: To build any crate using `everestrs`, this library must be findable. The only way I know to make it findable is by installing it, that seems the EVerest way to go. If you have a better suggestion, I would be happy to try it out, but likely in a followup MR.

everestrs/Cargo.lock
SirVer: yes, that is usually a good approach, even for a library. I guarantees reproducible builds. More is written here: https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control

everestrs/Cargo.toml
SirVer: It is common practice when you have C and Rust code on one level to have the build dir (`target`) on the top level.

everestrs/everestrs/build.rs
SirVer: As mentioned in our call yesterday, I think this is the biggest problem in the current approach: how to marry cargo & cmake. It is not ok to assume that cmake always drives the build process for rust, since this breaks all common development flows for rust engineers. I agree that this approach is brittle and I think we should find a better way to do it, but just assuming that cmake is always running the show is not working.

I suggest we spend time on this, some experimentation and building real modules will hopefully help, but right now I have no clear idea how to go about this.

everestrs/everestrs/src/lib.rs
SirVer: We have found another way - the void pointer is still there, but under the hood in the generated code by cxx.rs.

everestrs/everestrs/src/lib.rs
SirVer: I do not understand... a type alias for which type?

everestrs/CMakeLists.txt
SebaLukas: I would leave this in for now, was meant to be a quick fix here. I am probably not the only one with the problem.

everestrs/everestrs/src/lib.rs
SirVer: We found another way to hide the void pointer in the cxx.rs glue code.

everestrs/CMakeLists.txt
SirVer: ok

everestrs/everestrs/src/lib.rs
SirVer: my delightfully smart collegue @dorezyuk found a way of doing this in a safer manner by having c++ call back into Rust. This makes it harder to support multiple modules with one `Runtime` or `EverestHandle` for now, but this seems like a problem for later should it become one.

everestrs/everestrs/src/lib.rs
SirVer: no, it was not necessary, since we anyways only capture a pointer to the heap to the module, it is anyways always used with dynamic dispatch, so we can drop the templating. @dorezyuk pointed that out too. done now.

everestrs/everestrs/src/lib.rs
SirVer: My thought process here was the following: 

- Since the Rust module can do any prep before it initially EVerest, it is ready for EVerest as soon as it instantiates it.
- It is a non-goal for the Rust code to manually deal with fullfillments, this should all be transparently done by the framework. The flexibility that for example Python and JS is giving is currently explicitly not desired - it leaves too much room for human error. 
- It might or might not be interesting to the module to know when every module is ready, so I added the `on_ready` callback for it too know.

I might miss an important point here, but if so, I do not see it now. I added a TODO to discuss this at a later point in time.

everestrs/everestrs/src/lib.rs
SirVer: I do not understand your concerns with this. This is a pretty common pattern in Rust and it does not imbue runtime or compile cost (compared to static functions in C++ classes which need to be mentioned in headers for example).

everestrs/everestrs/src/lib.rs
SirVer: > I personally like this flexibility to create/implement various modules inside one unit for running tests and things like that.

I like that too, but I also find the danger of misuse high: instantiating twice the same module would yield strange behaviour, but is not technically impossible. Here it is. I can see the usefulness to bring up multiple modules in the same binary for testing, but I'd favor the approach to give each of them their own `Runtime` - or `EVerest handle`.

>  Concerning the link between both, why not having the rust module struct having it's one member variable, linking to the cxx module?

I could not parse that. What is it you are suggesting?

everestrs/everestrs/src/lib.rs
SirVer: I think this is a misread:

It is true that we pass the full schema as a serialized String - which is both wasteful and overkill - but we only deserialize from it what we need from it. To get what we need, we need to know the Schema anyways since we only have json in C++ land either. So either in Rust or in the C++ glue code in `everestrs_sys.cpp` we need to use the schema. I choose to do it in Rust and automatically deserialize it with a Framework because this is the safest and least error prone - and also very performant. Should the schema ever change, this problem will also much easier to find and fix than for example in the Python code, where the schema is used implicitly through dictionaries.

everestrs/everestrs/src/schema/interface.rs
SirVer: Once this lands we can switch to that - which I would prefer to do, but for now as pointed out above the framework leaves me no choice.

a-w50 (approved):

SirVer:
Thanks for the approval. I have no rights to merge this though, who can help me with this?

hikinggrass:
> Thanks for the approval. I have no rights to merge this though, who can help me with this?

I can merge this, one minor thing would be to run clang-format over the c++ code, we do have a dockerized version of the exact version we require here, but you can see the output as well in the Lint step of the github action.

One example command line would be: (maybe adding --user to provider your correct userid/groupid)
```bash
docker run --volume "$(pwd):/source" [ghcr.io/everest/everest-clang-format:latest](http://ghcr.io/everest/everest-clang-format:latest) -r /source --extensions "hpp,cpp" -e "/source/cache" -e "/source/build" -e "/source/build-cross" -e "/source/build-clang" -i
```

SirVer:
@hikinggrass Thanks for doing the formatting for me. I tried myself and the container does not run for me (MacOS Arm64):

~~~
docker run --volume "$(pwd):/source" ghcr.io/everest/everest-clang-format:latest -r /source --extensions "hpp,cpp" -e "/source/cache" -e "/source/build" -e "/source/build-cross" -e "/source/build-clang" -i
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
run-clang-format: error: Command '['clang-format', '--version']' returned non-zero exit status 255.
~~~

It seems the linking inside the container is not done correctly, though I am confused by the `qemu-x68_64`

hikinggrass:
> @hikinggrass Thanks for doing the formatting for me. I tried myself and the container does not run for me (MacOS Arm64):
> 
> ```
> docker run --volume "$(pwd):/source" ghcr.io/everest/everest-clang-format:latest -r /source --extensions "hpp,cpp" -e "/source/cache" -e "/source/build" -e "/source/build-cross" -e "/source/build-clang" -i
> qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
> run-clang-format: error: Command '['clang-format', '--version']' returned non-zero exit status 255.
> ```
> 
> It seems the linking inside the container is not done correctly, though I am confused by the `qemu-x68_64`

Ah that actually makes sense, since our container was initially just meant for the github action it just uses a prebuilt clang-format binary for x86-64. That then obviously doesn't work on arm64

PR #100, opened 2023-09-15
Allow setting user for all modules and capabilities per module

corneliusclaussen:
When setting system_user="myuser" property under settings in a config file all modules are started as this user and its default gid.

After starting all modules the manager also switches to this user.

For each module, special Linux capabilities can be specified with e.g. capabilities: "cap_chown+ep"
in the module section.

The manager will need to start as root or sufficient capabilities to do that. If no user/capabilities are specified no change happens.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,13 @@ find_package(Boost
     REQUIRED
 )
 
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(libcap
+    REQUIRED
+    IMPORTED_TARGET
+    libcap
+)
+
 if(NOT DISABLE_EDM)
     evc_setup_edm()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14)
 
 project(everest-framework
-    VERSION 0.8.1
+    VERSION 0.9
     DESCRIPTION "The open operating system for e-mobility charging stations"
 	LANGUAGES CXX C
 )
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -44,49 +44,6 @@ const auto PARENT_DIED_SIGNAL = SIGTERM;
 const int CONTROLLER_IPC_READ_TIMEOUT_MS = 50;
 auto complete_start_time = std::chrono::system_clock::now();
 
-class SubprocessHandle {
-public:
-    bool is_child() const {
-        return this->pid == 0;
-    }
-
-    void send_error_and_exit(const std::string& message) {
-        // FIXME (aw): howto do asserts?
-        assert(pid == 0);
-
-        write(fd, message.c_str(), std::min(message.size(), MAX_PIPE_MESSAGE_SIZE - 1));
-        close(fd);
-        _exit(EXIT_FAILURE);
-    }
-
-    // FIXME (aw): this function should be callable only once
-    pid_t check_child_executed() {
-        assert(pid != 0);
-
-        std::string message(MAX_PIPE_MESSAGE_SIZE, 0);
-
-        auto retval = read(fd, message.data(), MAX_PIPE_MESSAGE_SIZE);
-        if (retval == -1) {
-            throw std::runtime_error(fmt::format(
-                "Failed to communicate via pipe with forked child process. Syscall to read() failed ({}), exiting",
-                strerror(errno)));
-        } else if (retval > 0) {
-            throw std::runtime_error(fmt::format("Forked child process did not complete exec():\n{}", message.c_str()));
-        }
-
-        close(fd);
-        return pid;
-    }
-
-private:
-    const size_t MAX_PIPE_MESSAGE_SIZE = 1024;
-    SubprocessHandle(int fd, pid_t pid) : fd(fd), pid(pid){};
-    int fd{};
-    pid_t pid{};
-
-    friend SubprocessHandle create_subprocess(bool set_pdeathsig);
-};
-
 #ifdef ENABLE_ADMIN_PANEL
 class ControllerHandle {
 public:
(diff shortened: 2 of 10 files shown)


corneliusclaussen:
one thing that we also need to fix: if one child cannot be forked (e.g. capabilities config wrong), manager exits but the already forked children are not killed

a-w50:
that is, what `PR_SET_PDEATHSIG` in the SubProcess class is for, but somehow, this doesn't always work, as it seems

hikinggrass:
> that is, what `PR_SET_PDEATHSIG` in the SubProcess class is for, but somehow, this doesn't always work, as it seems

I think that's because that flag is cleared when executing setuid, so we will have to implement a workaround along these lines:
https://stackoverflow.com/a/42498370

a-w50:
> > that is, what `PR_SET_PDEATHSIG` in the SubProcess class is for, but somehow, this doesn't always work, as it seems
> 
> I think that's because that flag is cleared when executing setuid, so we will have to implement a workaround along these lines: https://stackoverflow.com/a/42498370

Good find, I wasn't aware of that yet.

golovasteek:
Hi in this PR quite nice refactoring is done, and all system-specific functions are encapsulated in a dedicated file.

Do you think `system_linux.cpp` would be more appropriate name for it, since the functionality there is rather linux specific and do not work unix-wide (macos, FreeBSD).

Do you think it makes sense to have one header `system.hpp` that declares all the system-dependent functions. And set of implementations e.g. here would be `system_linux.cpp` later `system_macos.cpp` could be added.

corneliusclaussen:
PR_SET_PDEATHSIG is now fixed, it is set after the user change. Also support to switch back to root user in manager for restarting of modules when admin panel/controller is used was added. If admin panel is disabled (which should be done on real deployments) manager does not have the caps to switch back from unprivileged user.

CMakeLists.txt
```diff
@@ -32,6 +32,13 @@ find_package(Boost
     REQUIRED
 )
 
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(libcap
```
hikinggrass, line 36: I think this dependency needs to be added to the readmes (at least for everest.github.io, maybe in everest-framework as well. Probably everest-core as well) Just so we don't forget :)

hikinggrass: should be libcap-dev on debian and libcap-devel on opensuse

src/manager.cpp
```diff
@@ -299,23 +227,21 @@ static std::map<pid_t, std::string> spawn_modules(const std::vector<ModuleStartI
...
-                break;
+        auto proc_handle = system::SubProcess::create(rs->run_as_user, module.capabilities);
+
+        if (proc_handle.is_child()) {
+            // first, check if we need any capabilities
+
+            try {
+                exec_module(rs, module, proc_handle);
+            } catch (const std::exception& err) {
+                // FIXME (aw): seems like other processes don't get killed, when this happens
```
hikinggrass: Is this FIXME still relevant?

src/system_unix.cpp
```diff
@@ -0,0 +1,214 @@
...
+    std::vector<gid_t> groups;
+
+    operator bool() const {
+        return this->error.empty();
+    }
+};
+
+static GetPasswdEntryResult get_passwd_entry(const std::string& user_name) {
+    // Assuming that a user does not have more than 50 groups
+    constexpr int max_supplementary_groups = 50;
```
hikinggrass, line 44: That's probably not relevant "in the real world" but the real limit is certainly much higher :D

CMakeLists.txt
hikinggrass, line 1: since this is a change which introduces new functionality we could bump the VERSION to 0.9.0 and tag appropriately afterwards

src/system_unix.cpp
```diff
@@ -0,0 +1,214 @@
...
+    for (const auto& cap_name : capabilities) {
+        auto& cap_value = capability_values.emplace_back();
+        const auto error = cap_from_name(cap_name.c_str(), &cap_value);
+
+        if (error) {
+            return fmt::format("Failed to get capability value for capability name {}", cap_name);
+        }
+    }
+
+    // FIXME (aw): error handling!
```
hikinggrass: Should this be addressed now or in a later PR?

src/system_unix.cpp
```diff
@@ -0,0 +1,214 @@
...
+    const auto set_uid_failed = setuid(entry.uid);
+    if (set_uid_failed) {
+        return "setuid failed";
+    }
+
+    return {};
+}
+
+void SubProcess::send_error_and_exit(const std::string& message) {
+    // FIXME (aw): howto do asserts?
```
hikinggrass: This will fail when pid == 0 is false, otherwise pass. I think that's correct here

src/system_unix.cpp
```diff
@@ -0,0 +1,214 @@
...
+    assert(pid == 0);
+
+    write(fd, message.c_str(), std::min(message.size(), MAX_PIPE_MESSAGE_SIZE - 1));
+    close(fd);
+    _exit(EXIT_FAILURE);
+}
+
+// FIXME (aw): this function should be callable only once
```
hikinggrass: maybe a job for std::call_once somewhere?

CMakeLists.txt
corneliusclaussen: I added them to the core readme which builds into everest.github.io. Readme of this repo does not contain any deps as far as I can see?

hikinggrass (changes requested):

src/system_unix.cpp
```diff
@@ -170,28 +170,31 @@ SubProcess SubProcess::create(const std::string& run_as_user, const std::vector<
...
+            auto error = system::set_real_user(run_as_user);
             if (not error.empty()) {
-                throw std::runtime_error(fmt::format("Failed to set capabilities: {}", error));
+                throw std::runtime_error(fmt::format("Failed to set real user to: {}", run_as_user));
             }
-        }
 
+            // Set capabilities for child process
+            if (not capabilities.empty()) {
+                error = system::set_caps(capabilities);
```
a-w50: we could just move this block into a separate function (would make things easier), and also, instead of throwing, we could again just return an error string

src/manager.cpp
corneliusclaussen: no, removed

src/system_unix.cpp
corneliusclaussen: agreed, but no idea what a good limit is here

CMakeLists.txt
corneliusclaussen: yes

src/system_unix.cpp
corneliusclaussen: error handling added

src/system_unix.cpp
corneliusclaussen: done, removed the try/catch

src/system_unix.cpp
corneliusclaussen: added a bool flag that allows the function to be called multiple times

src/system_unix.cpp
corneliusclaussen: removed the fixme

hikinggrass (approved):

PR #101, opened 2023-09-28
Make controller RPC timeout configurable

hikinggrass:
Also make it available to the admin panel

### DIFF
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -240,6 +240,13 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string& 
         controller_port = defaults::CONTROLLER_PORT;
     }
 
+    const auto settings_controller_rpc_timeout_ms_it = settings.find("controller_rpc_timeout_ms");
+    if (settings_controller_rpc_timeout_ms_it != settings.end()) {
+        controller_rpc_timeout_ms = settings_controller_rpc_timeout_ms_it->get<int>();
+    } else {
+        controller_rpc_timeout_ms = defaults::CONTROLLER_RPC_TIMEOUT_MS;
+    }
+
     const auto settings_mqtt_broker_host_it = settings.find("mqtt_broker_host");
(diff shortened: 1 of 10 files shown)


lib/runtime.cpp
```diff
@@ -240,6 +240,13 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string&
         controller_port = defaults::CONTROLLER_PORT;
     }
 
+    const auto settings_controller_rpc_timeout_ms_it = settings.find("controller_rpc_timeout_ms");
+    if (settings_controller_rpc_timeout_ms_it != settings.end()) {
+        controller_rpc_timeout_ms = settings_controller_rpc_timeout_ms_it->get<int>();
+    } else {
+        controller_rpc_timeout_ms = defaults::CONTROLLER_RPC_TIMEOUT_MS;
+    }
+
```
a-w50, line 249: how about
`controller_rpc_timeout_ms = settings.value("controller_rpc_timeout_ms", defaults::CONTROLLER_RPC_TIMEOUT_MS);`?

src/controller/rpc.cpp
```diff
@@ -91,6 +91,7 @@ class RPCRequest {
 
 RPC::RPC(int ipc_fd, const CommandApi::Config& config) : ipc_fd(ipc_fd) {
     this->api = std::make_unique<CommandApi>(config, *this);
+    this->rpc_timeout = std::chrono::milliseconds(config.controller_rpc_timeout_ms);
 }
```
a-w50: using the member initialization list (`ipc_fd(ipc_fd), rpc_timeout(std::chrono::milliseconds(config.controller_rpc_timeout_ms))` should be best practice

a-w50 (approved):

lib/runtime.cpp
hikinggrass: good idea, right now it just follows the style of the surrounding code, we should probably refactor this in a different PR though

PR #102, opened 2023-10-10
Rust integration: Adding variables & clients.

SirVer:
Co-authored-by: Dima Dorezyuk <ddo@qwello.eu>
Signed-off-by: Holger Rapp <hra@qwello.eu>


SirVer:
@hikinggrass Kai, could you run clang-format for me again, please? I still do not have access to a Linux dev machine.

Update: Figured it out.

SirVer:
We updated also https://github.com/EVerest/everest-core/pull/344 to work with this new code.

SirVer:
@hikinggrass @a-w50 Friendly ping? This has been sitting for a week now.

hikinggrass:
> @hikinggrass @a-w50 Friendly ping? This has been sitting for a week now.

I'm still working on the cmake integration, but I hope to have something to push sometime today

hikinggrass:
> > @hikinggrass @a-w50 Friendly ping? This has been sitting for a week now.
> 
> I'm still working on the cmake integration, but I hope to have something to push sometime today

This (in combination with the changes I just pushed on everest-core) should allow rust modules to be built with cmake without having to manually specify the location of everest-framework or to rely on heuristics. There have been some changes to the config handling in everest-framework recently which I've already rebased against.
The cmake support isn't completely ready yet and still needs some adjusting before merging.

One thing I've noticed with the rust modules in general: it looks like they do not yet publish true to everest/<module_id>/ready once their initialization code has run and triggering this manually doesn't seem to call the on_ready handler yet. That's probably something we should fix before merging this as well.

everestrs/everestrs/src/lib.rs
```diff
@@ -104,73 +123,147 @@ pub trait GenericModule: Sync {
...
+        Self {
             cpp_module,
-            module_impl: Box::pin(module_impl),
-        };
+            sub_impl: RwLock::new(None),
+        }
+    }
+
+    pub fn set_subscriber(&self, sub_impl: Weak<dyn Subscriber>) {
+        *self.sub_impl.write().unwrap() = Some(sub_impl);
```
a-w50: from an api perspective, `set_subscriber` can only be called once, right?  If so, this should belong into the constructor.

everestrs/everestrs/src/lib.rs
```diff
@@ -104,73 +123,147 @@ pub trait GenericModule: Sync {
...
             let interface: schema::Interface = interface_s.deserialize();
             for (name, _) in interface.cmds {
-                let meta = ffi::CommandMeta {
-                    implementation_id: implementation_id.clone(),
-                    name,
-                };
+                self.cpp_module
+                    .as_ref()
+                    .unwrap()
+                    .provide_command(self, implementation_id.clone(), name);
```
a-w50: Question: The runtime (self) is passed here via `provide_command`.  Are rust references direct pointers, i.e. do they point to a specific memory location?  If so, this makes the Runtime class non-moveable, correct?

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -57,13 +57,34 @@ void Module::signal_ready(const Runtime& rt) const {
...
-        JsonBlob blob = rt.handle_command(meta, json2blob(args));
+void Module::provide_command(const Runtime& rt, rust::String implementation_id, rust::String name) const {
+    handle_->provide_cmd(std::string(implementation_id), std::string(name), [&rt, implementation_id, name](json args) {
+        JsonBlob blob = rt.handle_command(implementation_id, name, json2blob(args));
         return json::parse(blob.data.begin(), blob.data.end());
     });
 }
 
+void Module::subscribe_variable(const Runtime& rt, rust::String implementation_id, rust::String name) const {
+    // TODO(hrapp): I am not sure how to model the multiple slots that could theoretically be here.
```
a-w50, line 68: @SirVer : what do you mean by multiple slots?

everestrs/everestrs/build.rs
```diff
@@ -7,6 +7,10 @@ struct Libraries {
 }
 
 fn find_everest_workspace_root() -> PathBuf {
+    if let Ok(everest_framework_dir) = env::var("EVEREST_RS_FRAMEWORK_SOURCE_LOCATION") {
```
a-w50, line 10: The naming is a bit off here, is it everest framework dir OR everestrs framework source location?

everestrs/everestrs/build.rs
```diff
@@ -24,8 +28,19 @@ fn find_everest_workspace_root() -> PathBuf {
...
+            (
+                everest_framework_path.join("everestrs/libeverestrs_sys.a"),
+                everest_framework_path.join("lib/libframework.so"),
+            )
+        } else {
+            (
+                root.join("everest-framework/build/everestrs/libeverestrs_sys.a"),
+                root.join("everest-framework/build/lib/libframework.so"),
+            )
+        };
```
a-w50, line 43: Not sure about rust best practices, but for me this `if let Ok(..) = ... {` doesn't naturally read.  Also the implicit returns are hard to spot (but could just be me ...)

everestrs/everestrs/build.rs
```diff
@@ -24,8 +28,19 @@ fn find_everest_workspace_root() -> PathBuf {
 /// Returns the Libraries path if this is a standalone build of everest-framework or None if it is
 /// not.
 fn find_libs_in_everest_framework(root: &Path) -> Option<Libraries> {
-    let everestrs_sys = root.join("everest-framework/build/everestrs/libeverestrs_sys.a");
-    let framework = root.join("everest-framework/build/lib/libframework.so");
+    let (everestrs_sys, framework) =
+        if let Ok(everest_framework_dir) = env::var("EVEREST_RS_FRAMEWORK_BINARY_LOCATION") {
```
a-w50, line 32: Compare the naming:
`if let Ok(everest_framework_dir) = env::var("EVEREST_RS_FRAMEWORK_SOURCE_LOCATION") {`
and
`if let Ok(everest_framework_dir) = env::var("EVEREST_RS_FRAMEWORK_BINARY_LOCATION") {`

a-w50, line 32: If you pass locations by environment variables, why don't you pass the the complete path to `libeverestrs_sys.a` and `libframework.so` directly, this would be more stable.  Note that at least inside cmake you can get the full path to library-targets (using generator expressions: `FRAMEWORK_LIB=$<TARGET_FILE:everest::framework>`)

src/manager.cpp
```diff
@@ -199,6 +200,26 @@ static SubprocessHandle exec_cpp_module(const ModuleStartInfo& module_info, std:
     return handle;
 }
 
+static SubprocessHandle exec_rust_module(const ModuleStartInfo& module_info, std::shared_ptr<RuntimeSettings> rs) {
```
a-w50: There is really no point here in passing the `RuntimeSettings` as shared pointer, just pass it as a (ideally const) reference.

src/manager.cpp
```diff
@@ -392,14 +415,31 @@ static std::map<pid_t, std::string> start_modules(Config& config, MQTTAbstractio
...
+                if (isupper(character)) {
+                    rust_filename += tolower(character);
+
+                } else {
+                    rust_filename += character;
+                }
+            }
+            char_count++;
+        }
+        auto rust_binary_path = module_path / "bin" / rust_filename;
```
a-w50: Is this really necessary? I hope not. If so, this should be put into a helper function with a proper name to improve readabilitiy

a-w50 (dismissed):

everestrs/everestrs/build.rs
hikinggrass: My intention with naming here was that this environment variable is meant for everestrs and contains the location to the source of everest-framework

everestrs/everestrs/build.rs
hikinggrass: we do need the binary location here, not the source location, which are usually not identical

hikinggrass: Absolutely, would add yet another environment variable but I guess the cost of that is quite low

src/manager.cpp
hikinggrass: At the moment this is just to keep it in line with the other exec_* functions, I would suggest we refactor this in a new PR

src/manager.cpp
hikinggrass: I admit this is the biggest hack here, the easiest would be to just start naming the rust binary targets in CamelCase as well, then we wouldn't need this hacky conversion in the first place. But I'm not sure about the rust conventions here, would this be alright @SirVer ?

everestrs/everestrs/src/lib.rs
dorezyuk: The better place would be the ctor but for our user facing api we have following dependencies 

- The publishers need a ref to the runtime
- the subscribers need a ref to the publishers
- The runtime needs a ref to the subscribers.

to brake this vicious cycle we pass the subscribers after the construction to the runtime

have a look at the core or for how it works together

everestrs/everestrs/src/lib.rs
dorezyuk: This is a good point. But we have to pin not the subscriber reference but the runtime. The cxx bridge and set_subscribers should probably take Pin<Runtime> since this is the point from which we aren’t allowed to move. By not implementing Unpin we make sure that the runtime remains pinned once the set_subscriber is called

everestrs/everestrs_sys/everestrs_sys.cpp
SirVer: @a-w50 I mean that `num_connections  != 1` for this `implementation_id`. I did not understand the C++ code around this, hence I did not model it in Rust at all currently.  My understanding is that the c++ code initially assumed that there would only be one connection, but then this was extendended to allow for more than one connection. To me this is a topic that needs addressing, but can be done at a later point. I also need a clinic where somebody explains this to me (or @dorezyuk, whoever implements this for rust).

everestrs/everestrs/build.rs
SirVer: nit: I'd prefer a match statement here too. The implicit returns are rustic - i.e. idiomatic. They are like meaningful whitespace in Python - one first finds them appalling, once used to them they become rather nice.

src/manager.cpp
SirVer: Add a TODO then for now so we do not forget? Reference this MR for posterity.

SirVer:
I did a flyby of the code, I have little time this week, so might not be able to dive deeper - thanks for the work and review so far!

hikinggrass (approved):

SirVer:
@hikinggrass What is this still waiting on?

hikinggrass:
> @hikinggrass What is this still waiting on?

just a rebase, I'm squashing right now :)

SirVer:
❤️ ❤️ 🎆 H O O R A Y  🚀 ❤️ ❤️

PR #113, opened 2023-11-07
Rust: Add support for the config

dorezyuk:
The PR adds rust support for the configuration handling. 
The PR goes hand in hand with the pr into everest-core https://github.com/EVerest/everest-core/pull/418

### DIFF
--- a/everestrs/everestrs/src/lib.rs
+++ b/everestrs/everestrs/src/lib.rs
@@ -3,6 +3,7 @@ mod schema;
 use argh::FromArgs;
 use serde::de::DeserializeOwned;
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use std::path::PathBuf;
 use std::pin::Pin;
 use std::sync::Arc;
@@ -36,6 +37,52 @@ mod ffi {
 
     struct JsonBlob {
         data: Vec<u8>,
+    }
+
+    /// The possible types a config can have. Note: Naturally this would be am
+    /// enum **with** values - however, cxx can't (for now) map Rusts enums to
+    /// std::variant or union.
+    enum ConfigType {
+        Boolean = 0,
+        String = 1,
+        Number = 2,
+        Integer = 3,
+    }
+
+    /// One config entry: As said above, we can't use an enum and have to
+    /// declare all values. Note: also Option is not an option...
+    struct ConfigField {
+        /// The name of the option, e.x. `max_voltage`.
+        name: String,
+
+        /// Our poor-mans enum.
+        config_type: ConfigType,
+
+        /// The value of the config field. The field has only a meaning if
+        /// `conifg_type is set to [ConfigType::Boolean].
+        bool_value: bool,
+
+        /// The value of the config field. The field has only a meaning if
+        /// `config_type` is set to [ConfigType::String].
+        string_value: String,
+
+        /// The value of the config field. The field has only a meaning if
+        /// `config_type` is set to [ConfigType::Number].
+        number_value: f64,
+
+        /// The value of the config field. The field has only a meaning if
+        /// `config_type` is set to [ConfigType::Integer].
+        integer_value: i64,
+    }
+
+    /// The configs of one module. Roughly maps to the cpp's counterpart
+    /// `ModuleConfig`.
+    struct RsModuleConfig {
+        /// The name of the group, e.x. "PowerMeter".
+        module_name: String,
+
+        /// All `ConfigFields` in this group.
+        data: Vec<ConfigField>,
     }
 
     unsafe extern "C++" {
@@ -85,8 +132,14 @@ mod ffi {
 
         /// Publishes the given `blob` under the `implementation_id` and `name`.
         fn publish_variable(self: &Module, implementation_id: &str, name: &str, blob: JsonBlob);
+
+        /// Returns the module config from cpp.
+        fn get_module_configs(module_id: &str, prefix: &str, conf: &str) -> Vec<RsModuleConfig>;
+
     }
 }
+
+// type RsModuleConfigs = HashMap<String, HashMap<String, ConfigTypes>>;
 
 impl ffi::JsonBlob {
     fn as_bytes(&self) -> &[u8] {
@@ -286,4 +339,87 @@ impl Runtime {
         // again.
         (self.cpp_module).as_ref().unwrap().signal_ready(self);
     }
+}
+
+/// A store for our config values. The type is closely related to
+/// [ffi::ConfigField] and [ffi::ConfigType].
+#[derive(Debug)]
+pub enum Config {
+    Boolean(bool),
+    String(String),
+    Number(f64),
+    Integer(i64),
+}
+
+impl TryFrom<&Config> for bool {
+    type Error = ();
+    fn try_from(value: &Config) -> std::result::Result<Self, Self::Error> {
+        match value {
+            Config::Boolean(value) => Ok(*value),
+            _ => Err(()),
+        }
+    }
+}
+
+impl TryFrom<&Config> for String {
+    type Error = ();
+    fn try_from(value: &Config) -> std::result::Result<Self, Self::Error> {
+        match value {
+            Config::String(value) => Ok(value.clone()),
+            _ => Err(()),
+        }
+    }
+}
+
+impl TryFrom<&Config> for f64 {
+    type Error = ();
+    fn try_from(value: &Config) -> std::result::Result<Self, Self::Error> {
+        match value {
+            Config::Number(value) => Ok(*value),
+            _ => Err(()),
+        }
+    }
+}
+
+impl TryFrom<&Config> for i64 {
+    type Error = ();
+    fn try_from(value: &Config) -> std::result::Result<Self, Self::Error> {
+        match value {
+            Config::Integer(value) => Ok(*value),
+            _ => Err(()),
+        }
+    }
+}
+
+/// Interface for fetching the configurations through the C++ runtime.
+pub fn get_module_configs() -> HashMap<String, HashMap<String, Config>> {
+    let args: Args = argh::from_env();
+    let raw_config = ffi::get_module_configs(
+        &args.module,
+        &args.prefix.to_string_lossy(),
+        &args.conf.to_string_lossy(),
+    );
+
+    // Map the nested Vec's into a nested HashMaps.
+    raw_config
+        .into_iter()
+        .map(|ffi::RsModuleConfig { module_name, data }| {
+            let current = data
+                .into_iter()
+                .map(|field| {
+                    let value = match field.config_type {
+                        ffi::ConfigType::Boolean => Config::Boolean(field.bool_value),
+                        ffi::ConfigType::String => Config::String(field.string_value),
+                        ffi::ConfigType::Number => Config::Number(field.number_value),
+                        ffi::ConfigType::Integer => Config::Integer(field.integer_value),
+                        _ => panic!("This should never happen"),
+                    };
+
+                    (field.name, value)
+                })
+                .collect::<HashMap<_, _>>();
+
+            (module_name, current)
+        })
+        .collect::<HashMap<_, _>>()
 }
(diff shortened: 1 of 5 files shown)


everestrs/everestrs/src/lib.rs
```diff
@@ -38,6 +39,52 @@ mod ffi {
         data: Vec<u8>,
     }
 
+    /// The possible types a config can have. Note: Naturally this would be am
+    /// enum **with** values - however, cxx can't (for now) map Rusts enums to
+    /// std::variant or union.
+    enum ConfigType {
```
SirVer, line 48: use `serde_codegen::schema::interface::Value` and use Serde do deserialize stuff. That makes this code go away and the other code can be reused.

SirVer: This means pulling this into everstrs which seems fine.

everestrs/everestrs/src/lib.rs
```diff
@@ -85,9 +132,15 @@ mod ffi {
...
         /// Publishes the given `blob` under the `implementation_id` and `name`.
         fn publish_variable(self: &Module, implementation_id: &str, name: &str, blob: JsonBlob);
+
+        /// Returns the module config from cpp.
+        fn get_module_configs(module_id: &str, prefix: &str, conf: &str) -> Vec<RsModuleConfig>;
+
     }
 }
 
+// type RsModuleConfigs = HashMap<String, HashMap<String, ConfigTypes>>;
```
SirVer: delete commented code?

everestrs/everestrs/src/lib.rs
```diff
@@ -287,3 +340,86 @@ impl Runtime {
...
+    fn try_from(value: &Config) -> std::result::Result<Self, Self::Error> {
+        match value {
+            Config::Integer(value) => Ok(*value),
+            _ => Err(()),
+        }
+    }
+}
+
+/// Interface for fetching the configurations through the C++ runtime.
+pub fn get_module_configs() -> HashMap<String, HashMap<String, Config>> {
```
SirVer, line 396: return the Json blob and deseriailze it like we do with everything else. It is not worth to have two different ways of handling deserialization in this module. And as mentioned, the code is in the codegen branch already.

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -80,11 +81,56 @@ JsonBlob Module::call_command(rust::Str implementation_id, rust::Str name, JsonB
     return json2blob(return_value);
 }
 
+void Module::publish_variable(rust::Str implementation_id, rust::Str name, JsonBlob blob) const {
```
SirVer, line 121: why was this moved here?

everestrs/everestrs/src/lib.rs
dorezyuk: Are you suggesting to serialize the stuff in the c++ land and then pass json to rust and then deserialize it again?

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -80,11 +81,56 @@ JsonBlob Module::call_command(rust::Str implementation_id, rust::Str name, JsonB
...
+}
+
 std::unique_ptr<Module> create_module(rust::Str module_id, rust::Str prefix, rust::Str conf) {
     return std::make_unique<Module>(std::string(module_id), std::string(prefix), std::string(conf));
 }
 
-void Module::publish_variable(rust::Str implementation_id, rust::Str name, JsonBlob blob) const {
-    handle_->publish_var(std::string(implementation_id), std::string(name),
-                         json::parse(blob.data.begin(), blob.data.end()));
+namespace {
```
SirVer: there is already an anonymous namespace at the top of the file. Just keep one please.

everestrs/everestrs_sys/everestrs_sys.hpp
```diff
@@ -4,11 +4,16 @@
 #include <framework/runtime.hpp>
 #include <memory>
 #include <string>
+#include <utils/types.hpp>
 
 #include "cxxbridge/rust.h"
+// #include "cxxbridge/lib.rs.h"
```
SirVer: remove commented code?

everestrs/everestrs/src/lib.rs
SirVer: actually I misunderstood the C++ stuff - I was assuming that json values are passed around, but for Config that is not the case. retracted.

everestrs/everestrs_sys/everestrs_sys.cpp
dorezyuk: To keep all impls for Module's functions together

everestrs/everestrs/src/lib.rs
```diff
@@ -287,3 +340,86 @@ impl Runtime {
...
+/// [ffi::ConfigField] and [ffi::ConfigType].
+#[derive(Debug)]
+pub enum Config {
+    Boolean(bool),
+    String(String),
+    Number(f64),
+    Integer(i64),
+}
+
+impl TryFrom<&Config> for bool {
```
SirVer, line 355: Why is this not using everestrs::Error? also below.

everestrs/everestrs/src/lib.rs
SirVer: retracted.

everestrs/everestrs/src/lib.rs
```diff
@@ -287,3 +340,86 @@ impl Runtime {
...
+        .map(|ffi::RsModuleConfig { module_name, data }| {
+            let current = data
+                .into_iter()
+                .map(|field| {
+                    let value = match field.config_type {
+                        ffi::ConfigType::Boolean => Config::Boolean(field.bool_value),
+                        ffi::ConfigType::String => Config::String(field.string_value),
+                        ffi::ConfigType::Number => Config::Number(field.number_value),
+                        ffi::ConfigType::Integer => Config::Integer(field.integer_value),
+                        _ => panic!("This should never happen"),
```
SirVer: not a helpful error message. Please include the unexpected value. Or use `unreachable!`.

everestrs/everestrs/src/lib.rs
```diff
@@ -287,3 +340,86 @@ impl Runtime {
...
+pub fn get_module_configs() -> HashMap<String, HashMap<String, Config>> {
+    let args: Args = argh::from_env();
+    let raw_config = ffi::get_module_configs(
+        &args.module,
+        &args.prefix.to_string_lossy(),
+        &args.conf.to_string_lossy(),
+    );
+
+    // Map the nested Vec's into a nested HashMaps.
+    raw_config
```
SirVer: this is very hard to read. Can you instead `push` into a named vector so the stuff is easier to read.

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -80,11 +81,56 @@ JsonBlob Module::call_command(rust::Str implementation_id, rust::Str name, JsonB
...
+
+inline ConfigField get_config_field(const std::string& _name, bool _value) {
```
SirVer: Very clever, but please add a comment what you are doing here.

everestrs/everestrs_sys/everestrs_sys.cpp
dorezyuk: I wanted to keep these functions close the the variant - so they are its immediate apparent why they are here. If you feel strange about 2 namespaces - we can remove it and make the functions static. But I feel the closer these functions are to the variant the easier its to maintain and understand

everestrs/everestrs/src/lib.rs
dorezyuk: It should never happen however if c++ casts an invalid value to enum class it should be UB and I'm not sure what then happens here - will print the value

everestrs/everestrs_sys/everestrs_sys.cpp
SirVer: I feel really juck about two anonymous namespaces in one file and static functions are even worse. I appreciate the argument of keeping these close to their use, but I still think it better to move them into the anonymous namespace and only have one area of private stuff in the file. Pretty please with sugar on top? 🐱

SirVer (approved):
just one more nit, otherwise looks good to me. This in fact looks really nice and seems quite future proof.

SirVer:
@hikinggrass @a-w50 If you still want to review this before we merge it, please speak up. Otherwise we'll probably merge it Friday or monday.

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -30,6 +33,41 @@ JsonBlob json2blob(const json& j) {
...
+template <typename T, typename... VARIANT_T> struct VariantMemberImpl : public std::false_type {};
+
+template <typename T, typename... VARIANT_T>
+struct VariantMemberImpl<T, std::variant<VARIANT_T...>> : public std::disjunction<std::is_same<T, VARIANT_T>...> {};
+
+/// @brief Static checker if the type T can be converted to `ConfigEntry`.
+///
+/// We use this to detect `get_config_field` overloads which receive arguments
+/// which aren't part of our `ConfigEntry` variant.
+template <typename T> struct ConfigEntryMember : public VariantMemberImpl<T, ConfigEntry> {};
```
a-w50, line 49: in line with the c++ standard, I would call the struct `is_config_entry_member` or something similar

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -30,6 +33,41 @@ JsonBlob json2blob(const json& j) {
...
+
+inline ConfigField get_config_field(const std::string& _name, bool _value) {
```
a-w50, line 51: i would probably have created a single class with visitor functions, which gets the name as its constructor argument and whose instance could then be used directly within `std::visit`

everestrs/everestrs_sys/everestrs_sys.cpp
a-w50: furthermore, if we're using `std::visit` we got already type checking, so if we don't implement the visitor for a type, it doesn't exist. Shouldn't be that enough?  Of course there might be still implicit conversions, but that might still happen here too

everestrs/everestrs_sys/everestrs_sys.cpp
```diff
@@ -80,11 +118,36 @@ JsonBlob Module::call_command(rust::Str implementation_id, rust::Str name, JsonB
...
+
+    // Iterate over all modules stored in the module_config.
+    for (const auto& mm : module_configs) {
+        RsModuleConfig mm_out{mm.first, {}};
+        mm_out.data.reserve(mm.second.size());
+
+        // Iterate over all configs stored in the mm (our current module).
+        for (const auto& cc : mm.second) {
+            mm_out.data.emplace_back(
+                std::visit([&](auto&& _value) { return ::get_config_field(cc.first, _value); }, cc.second));
```
a-w50, line 147: Unfortunately it is that painful to convert a c++ variant to a rust enum ...

include/utils/config.hpp
```diff
@@ -127,7 +127,7 @@ class Config {
     ///
     /// \brief checks if the config contains the given \p module_id
     ///
-    bool contains(const std::string& module_id);
+    bool contains(const std::string& module_id) const;
```
a-w50, line 130: :+1:

a-w50 (approved):
@SirVer @dorezyuk :+1: see my personal opinion on the visitor and the type trait.  Feel free to change it or not.

everestrs/everestrs_sys/everestrs_sys.cpp
dorezyuk: The everest projects are using CamelCase and not snake_case. 
As for why its there - it's primary task it to fail the compilation if some one removes an entry from std::variant - in this case the dev is "politely" hinted that there might be some cleanup to do in the rust-glue code

a-w50: I see, so in case we remove one type from the variant, one of the static asserts should fail, although they will never be called then anymore.  A single `std::is_same_v<ConfigEntry, std::variant<currently_known_types...>>` might also work here

everestrs/everestrs_sys/everestrs_sys.cpp
SirVer: not so surprising, since they have a different memory layout internally. I assume cxx.rs will eventually add code to make this easier, but for now it is what it is.

PR #114, opened 2023-11-09
Add Global Errors Feature

andistorm:


a-w50:
@andistorm concerning the tests refactoring, this shouldn't belong to *Add Global Errors Feature*, could you split this off to another PR?

lib/everest.cpp
```diff
@@ -422,6 +422,47 @@ void Everest::subscribe_error(const Requirement& req, const std::string& error_t
     this->mqtt_abstraction.register_handler(error_topic, token, QOS::QOS2);
 }
 
+void Everest::subscribe_all_errors(const JsonCallback& callback) {
```
hikinggrass, line 425: This is pretty much identical to the code in subscribe_all_errors_cleared, could this be refactored into a common function?

andistorm: Same is valid for `subscribe_error` and `subscribe_error_cleared`.
So I adapted the new feature to that.

But I agree it would make sense to reuse code here. I would refactor in another PR and go one step further:
1. Do `subscribe_error_cleared` in `subscribe_error`
2. Do `subscribe_all_errors_cleared` in `subscribe_all_errors`
Since the visible usage of this is already doing it (see auto gen code for cpp modules)

hikinggrass (approved):

PR #117, opened 2023-11-11
Make everest-framework compilable on macos

golovasteek:
## Objective
To develop modules on macos it is nice to have the sources that at least compile, so that different tools (IDEs) can use  them.
Compilable code enables the future work on making macos-based development of EVerest and modules. But it needs to be tested and may require some further work. So specific use-cases are not guaranteed to be working in this PR.

## Details

There were a couple of system calls, that are unavailable on macOS.

`prctl` - used only for the setting of the process name and some cleanup. Process name setting is just disabled, the cleanup maybe won't work. Needs review.
`pipe2` - on macos opening pipe, and setting flags are different calls, so it is now refactored a bit.

`libwebsockets` doesn't compile with openssl3 (AFAIK), so version 1.1 needs to be installed, and the paths should be set in CMake.

There is a workflow for macos build added, to ensure, that it still compiles on macos.

### DIFF
--- a/.github/workflows/macos-build.yaml
+++ b/.github/workflows/macos-build.yaml
@@ -26,6 +26,11 @@ jobs:
           path: everest-dev-environment
           ref: v0.5.5
 
+      - name: Cache build
+        uses: actions/cache@v3
+        with:
+          path: ${{github.workspace}}/build
+
       - name: Install dependencies with homebrew
         run: brew install openssl@1.1 boost
 
@@ -43,5 +48,6 @@ jobs:
             -DCMAKE_PREFIX_PATH=$EVEREST_CMAKE_PATH \
             -DLWS_OPENSSL_LIBRARIES="/usr/local/opt/openssl@1.1/lib/libcrypto.dylib;/usr/local/opt/openssl@1.1/lib/libssl.dylib" \
             -DLWS_OPENSSL_INCLUDE_DIRS="/usr/local/opt/openssl@1.1/include" \
-            -DEVEREST_ENABLE_JS_SUPPORT=OFF
+            -DEVEREST_ENABLE_JS_SUPPORT=OFF \
+            -DEVEREST_ENABLE_RS_SUPPORT=ON
           make -j2
--- a/.github/workflows/macos-build.yaml
+++ b/.github/workflows/macos-build.yaml
@@ -1,0 +1,47 @@
+name: MacOS Workflow
+on: [push]
+
+jobs:
+  build:
+    name: ensure that everest-framework builds on MacOS
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+        with:
+          path: everest-framework
+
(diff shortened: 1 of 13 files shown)


a-w50 (commented):
Thanks for that.  Nice to see, that it doesn't take that much modification, in order to get it build on mac os.
Some things I would change:
- workflow: if we aim for build/compile speed, I'd probably use caching and a dedicated docker container, which then would have openssl and boost already installed
- concerning prctl, I would factor this out into a separate cpp (maybe under directory called os, where later could all the os dependent stuff should go) and do the ifdefs there so the main application code stays "simple"
- concerning the pipe2 thing, same here
- `PR_SET_PDEATHSIG` results in sending a signal to the childs (modules) if the manager dies - so they get the possibility to shut down (their current default behavior).  I guess this should be possible in mac os somehow too

Thanks for your work :1st_place_medal:

corneliusclaussen:
This will probably break again with the merge of PR100 which should be finished soon. I like the idea to have things compilable on many different OS if possible, but I think we cannot make this mandatory requirement for pull requests, some commits will break mac os compile and that can be quite hard to fix if you do not have a mac.

golovasteek:
Would do you think would be the way forward?

I see two options:
 * keep the macOS compatibility as a best-effort. Keep the compatibility code but do not test in in CI.
 * declare EVerest as linux-only, and do not have anything related to macOS.

hikinggrass:
I would argue for a third option:
* Keep the MacOS compatibility as best effort, still build it in ci (so people without a Mac know they broke something) but do not require this ci run to pass

golovasteek:
It runs out that the same functions are refactored quite a bit in #100 , so it makes sense that #100 goes first.

hikinggrass:
> It runs out that the same functions are refactored quite a bit in #100 , so it makes sense that #100 goes first.

and we might solve one of the issues with this PR (prctl, especially PR_SET_PDEATHSIG not being directly available on macos) in #100 since we can't rely on it anymore when using setuid

a-w50:
> > It runs out that the same functions are refactored quite a bit in #100 , so it makes sense that #100 goes first.
> 
> and we might solve one of the issues with this PR (prctl, especially PR_SET_PDEATHSIG not being directly available on macos) in #100 since we can't rely on it anymore when using setuid

Doing some googling, it seems that there is not a simple solution like PR_SET_PDEATHSIG, which by itself has some pitfalls (like when using setuid ..). One way on MacOs seems to be, to listen/poll on a specific event (kqueue), but this would then need to be done by a separate thread (for every module/process).  Another option might be a simple neighboring process to the manager which is just there and checks if the manager crashed unexpectedly and would then shutdown the other modules.  This layout also seems to be the favored one in crashpad / breakpad (libraries for handling and logging unexpected crashes).  So we might choose to go this direction ...

corneliusclaussen:
Anyone still working on this PR?

golovasteek:
This is quite old, and I need to start over on the new state of the main branch.

Closing

PR #118, opened 2023-11-13
CI: build the everest-framework in the ci

dorezyuk:
We currently don't run any builds in this repository. This pr enables build and tests. However, the `test_config.cpp` are broken. Since there is a pr underway which fixes them, I've removed them from the CMakeLists.txt for now.

The pr enables by default Rust compilation.


.ci/build-kit/compile.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -e
+
+# Install coverage dependencies
+apt update && apt install -y gcovr lcov
```
dorezyuk, line 5: All the setup should go (in future) into the docker image

.github/workflows/linux-build.yaml
```diff
@@ -0,0 +1,43 @@
+name: Linux Build
+on: [workflow_dispatch, pull_request]
+
+jobs:
+  build:
+    name: Build
+    strategy:
+      matrix:
+        os: [large-ubuntu-22.04-xxl]
```
a-w50: We don't really need the paid fast runners, if caching is working properly.  So ubuntu-22.04 or what it is called, should be fine.  Furthermore, the matrix build is not necessary here, if we only use one platform anyway

.github/workflows/linux-build.yaml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Setup cache
+        uses: actions/cache@v3
+        with:
+          path: cache
+          key: compile-${{ github.ref }}-${{ github.sha }}
```
a-w50: Looks like this is based on the workflow from everest-core like two weeks ago.  The caching is not really working well here in cases where one has `-` in the branch name, which is not uncommon.  See the current workflow file in everest core on how to fix that

.ci/build-kit/compile.sh
```diff
@@ -0,0 +1,20 @@
...
+cmake \
+    -B build \
+    -S "$EXT_MOUNT/source" \
+    -G Ninja \
+    -DEVC_ENABLE_CCACHE=1 \
+    -DISO15118_2_GENERATE_AND_INSTALL_CERTIFICATES=OFF \
+    -DBUILD_TESTING=ON \
+    -DEVEREST_ENABLE_RS_SUPPORT=ON
+
+ninja -j$(nproc) -C build && ninja test -C build
```
a-w50, line 20: Just concerning the naming here: it is called compile, but actually it also runs the tests.  I would either split it into two or call it *compile and test* or something similar

.ci/build-kit/compile.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -e
+
+# Install coverage dependencies
+apt update && apt install -y gcovr lcov
+
+# Install Rust
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+source "$HOME/.cargo/env"
```
a-w50, line 9: I'm not sure on what the best solution here is.  Ideally all the build dependencies are already setup in the build container, but this would mean we either have one large container, which contains all the dependencies, or multiple (probably a lot) smaller ones, the contain only the dependencies needed.

a-w50:
@dorezyuk Thank you for your efforts.  I would like to have this CI pipeline setup for all repositories in everest.  I still haven't figured out, if the workflows can be made reusable in a way that make sense.  When I created the everest-ci repo and this "build-kit" stuff, I had the following points in mind:
- the github workflows should be easily reproducible on a dev system (therefor the docker containers from everest-ci and the build-kit shell scripts) without all the github specific stuff
- the github actions should run as fast as possible (using caching and prebuild containers), so we can use the free github runners
- everest-ci should serve as the basic infrastructure, that can be reused in all the everest repos, so code duplication can be kept small

If you're interested, we could have a separate call on this topic.

.ci/build-kit/compile.sh
dorezyuk: I think we can figure it out once we have more stability

andistorm:
@dorezyuk I hooked up your work here: https://github.com/EVerest/everest-framework/pull/136

andistorm:
Sorry for closing, I misclicked

andistorm:
Since https://github.com/EVerest/everest-framework/pull/136 is merged I close this one

PR #120, opened 2023-11-14
Disable redundant validations by default

hikinggrass:


lib/runtime.cpp
```diff
@@ -335,6 +335,13 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string&
     } else {
         validate_schema = defaults::VALIDATE_SCHEMA;
     }
+
+    const auto settings_validate_schema_on_startup_it = settings.find("validate_schema_on_startup");
+    if (settings_validate_schema_on_startup_it != settings.end()) {
+        validate_schema_on_startup = settings_validate_schema_on_startup_it->get<bool>();
+    } else {
+        validate_schema_on_startup = defaults::VALIDATE_SCHEMA_ON_STARTUP;
+    }
```
a-w50: I know the other code isn't doing it like that, but definitely, writing
`validate_schema_on_startup = settings.value("validate_schema_on_startup", defaults::VALIDATE_SCHEMA_ON_STARTUP);`
should be the correct choice here

include/framework/runtime.hpp
```diff
@@ -81,6 +81,7 @@ inline constexpr auto MQTT_EXTERNAL_PREFIX = "";
 inline constexpr auto TELEMETRY_PREFIX = "everest-telemetry";
 inline constexpr auto TELEMETRY_ENABLED = false;
 inline constexpr auto VALIDATE_SCHEMA = false;
+inline constexpr auto VALIDATE_SCHEMA_ON_STARTUP = false;
```
a-w50: For the user it might be hard to guess what the 'validate_schema' and 'validate_schema_on_startup' is doing.  So maybe we can find a better naming here.

lib/config.cpp
```diff
@@ -431,19 +431,22 @@ Config::Config(std::shared_ptr<RuntimeSettings> rs, bool manager) : rs(rs), mana
...
+                        auto end_time_validate = std::chrono::system_clock::now();
+                        EVLOG_debug << "YAML validation of " << types_entry.path().string() << " took: "
+                                    << std::chrono::duration_cast<std::chrono::milliseconds>(end_time_validate -
+                                                                                             start_time_validate)
+                                           .count()
+                                    << "ms";
+                        total_time_validation_ms += std::chrono::duration_cast<std::chrono::milliseconds>(
+                                                        end_time_validate - start_time_validate)
+                                                        .count();
+                    }
```
a-w50: Could we please move this block into a separate function? I could be reused below, as it does exactly the same.

lib/runtime.cpp
hikinggrass: Changed, this was really just done for consistency with the rest of the code.

include/framework/runtime.hpp
hikinggrass: Do you have a suggestion? I did not have a better idea

lib/config.cpp
hikinggrass: sure, refactored this part

lib/runtime.cpp
hikinggrass: removed

include/framework/runtime.hpp
hikinggrass: I removed this option altogether and only run this validation step if you set the existing validate_schema option in the config

a-w50 (approved):

PR #123, opened 2023-11-16
Fix missing Exception throw on loading config

andistorm:
In case of loading a broken config.yaml the function load_yaml returns a nullptr. This fix avoid accessing non-existing elements on a nullptr.

### DIFF
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -135,8 +135,8 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string& 
     if (config == nullptr) {
         EVLOG_info << "Config file is null, treating it as empty";
         config = json::object();
-    } else if (!config->is_object()) {
-        throw BootException(fmt::format("Config file '{}' is not an object", config_file));
+    } else if (!config.is_object()) {
+        throw BootException(fmt::format("Config file '{}' is not an object", config_file.string()));
     }
 
     const auto settings = config.value("settings", json::object());
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -132,6 +132,9 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string& 
     }
 
     config = load_yaml(config_file);
+    if (config == nullptr) {
+        EVLOG_AND_THROW(EverestConfigError("Config file is nullptr!"));
+    }
 
     const auto settings = config.value("settings", json::object());
 
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -133,7 +133,10 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string& 
 
     config = load_yaml(config_file);
     if (config == nullptr) {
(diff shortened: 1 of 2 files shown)


hikinggrass:
Could we move this check into load_yaml itself? @a-w50 we're using load_yaml in quite a few places in the code and I don't see any usage that would rely on returning a nullptr in case of a broken/empty yaml here. Or am I missing something?

lib/runtime.cpp
```diff
@@ -132,6 +132,9 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string&
     }
 
     config = load_yaml(config_file);
+    if (config == nullptr) {
+        EVLOG_AND_THROW(EverestConfigError("Config file is nullptr!"));
+    }
```
a-w50: I think `load_yaml` always returns an nlohmann::json instance but it can be implicitely equal to a `nullptr`, in case it is an empty file.  So this is not an intrinsic error and the message is also wrong in this case.  If we're not able to deal with "empty" config files (which I think, could be valid) then we should throw a corresponding error like 'empty config files not supported'

a-w50 (changes requested):

lib/runtime.cpp
andistorm: The problem doesn't seem to handle empty config files, but to handle broken config files. [This file](https://github.com/EVerest/everest-framework/blob/fix-unit-tests/tests/test_configs/broken_config_config.yaml) caused the error:
```
[json.exception.type_error.306] cannot use value() with null
```
I am also sure that the error occurs in [this line](https://github.com/EVerest/everest-framework/blob/9604e06820ea671b68d097870d2b80b6c99bfa1e/lib/runtime.cpp#L136)

Since there is a default value given it should'nt brake on empty configs (and it doesn't).
So it seems to make a difference laoding a broken config or an empty config

lib/yaml_loader.cpp
```diff
@@ -40,7 +40,7 @@ struct RymlCallbackInitializer {
 
 static nlohmann::ordered_json ryml_to_nlohmann_json(const c4::yml::NodeRef& ryml_node) {
     if (ryml_node.empty()) {
-        return nullptr;
+        return nlohmann::ordered_json::object();
```
andistorm: @a-w50 
I found the issue with testing with a ['broken' config](https://github.com/EVerest/everest-framework/blob/fix-unit-tests/tests/test_configs/broken_config_config.yaml), but it also seems to occur with empty config files as you said.

The error
```
[json.exception.type_error.306] cannot use value() with null
```
occured in [this line](https://github.com/EVerest/everest-framework/blob/9604e06820ea671b68d097870d2b80b6c99bfa1e/lib/runtime.cpp#L136).

Since there is a default value given it should'nt brake on empty objects.

With the mentioned fix it doesn't brake. Seems to make a difference to return nullptr or empty json here..


_In case of broken config (invalid yaml) it shouldn't return anything, but throw a exception, but I guess the 'broken' config I use is not broken enough_

a-w50 (approved):

hikinggrass (approved):

PR #125, opened 2023-11-17
Rust code generation

dorezyuk:
Rust code-generation support.

Co-authored-by: Holger Rapp [hra@qwello.eu](mailto:hra@qwello.eu)


a-w50:
Hi, right now, we still have the code generation in everest-utils, and we're planning to split it up between everest-framework and everest-core.  The idea was, that the part of the code generation, which needs to understand the schema definitions, will belong to everest-framework and the part which provides the language specific "frontend" will belong to everest-core.  As your implementation is only rust specific, this should PR should go into everest-core and not everest-framework.

dorezyuk:
> Hi, right now, we still have the code generation in everest-utils, and we're planning to split it up between everest-framework and everest-core. The idea was, that the part of the code generation, which needs to understand the schema definitions, will belong to everest-framework and the part which provides the language specific "frontend" will belong to everest-core. As your implementation is only rust specific, this should PR should go into everest-core and not everest-framework.

Is there some documentation where I can read up on this?

a-w50:
No, there is no documentation for that.  You can have a look at the branch `feature/ev-cli_refactor` in everest-core.  The code-generator script is inside `scripts/code-generator`.  There are two sub-folders
- core, which should be the part that will stay in core
- framework, the part which should go to everest-framework (for now I left it there so I don't need to handle two different repos)
From that you should get a good idea on how it should look like.  If you like we can also have a meeting and discuss that.

SirVer:
I think we have a couple of ways forward here:

1. Merge this code gen as is and later converge it towards the new design you are proposing, @a-w50.
2. Do not merge this code gen and try to guess the final design and converge it before merging.

I am very strongly in favor of 1, mostly because we want to build a product on this code and are pressed for time. I also do not agree that code-gen should be split between core and framework, I think it belongs in framework like proposed here. How are such design decisions figured out in the everest project?

everestrs/everestrs-build/jinja/config.jinja2
```diff
@@ -0,0 +1,48 @@
...
+    pub(crate) {{ config.name }}: {{ config.data_type }},
+    {% endfor %}
+
+    {% for p_config in provided_config %}
+    /// The config for the `{{ p_config.name }}` interface.
+    pub(crate) {{ p_config.name }}_config: {{ p_config.name | title }}Config,
+    {% endfor %}
+}
+
+/// Returns the config for the whole module. You can call this function at any
```
SirVer: nit: What does it mean that you can "call this function at any point in time"? Be more concrete, like "this can be called before Module instantiation" or " can be called before the Everest Runtime is constructed"

SirVer (approved):
only one comment remaining, looks good to me.

SirVer:
@dorezyuk Please make the MR description more detailed. also add co-author(s), at least me contributed. The MR description will become the squashed commit description.

everestrs/Cargo.toml
```diff
@@ -2,4 +2,8 @@
 resolver = "2"
 members = [
    "everestrs",
+   "everestrs-build",
 ]
+
+[workspace.dependencies]
+everestrs-build = { path="everestrs-build" }
```
hikinggrass, line 9: Is there a reason why everestrs depends on everestrs-build and not the other way around?

everestrs/everestrs/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-mod schema;
+use everestrs_build::schema;
```
hikinggrass, line 1: Looks like this is the reason for the dependency inversion. Wouldn't it be possible to keep the schema implementations in everestrs and use them from everestrs_build, or is there a technical reason that I missed that necessitates moving these into everestrs?

everestrs/Cargo.toml
dorezyuk: yes kind of - the build dependencies (read everestrs-build) are build before other dependencies (read everestrs)

SirVer:
@a-w50  My understanding is that you changed something on the branch, but I do not see it. And it might have have been pushed over during a rebase of @dorezyuk (if  so, sorry for that). I discussed with @caller and we decided to merge for now, let's discuss further refactorings in the tuesday call.

For the moment we also branched into https://github.com/qwello core and framework for now and will do further development there for higher development speed. I shortly discussed these topics with @caller, but we need to further discuss how collaboration can be done with less friction going forward.

PR #129, opened 2023-11-23
Construct explanatory string of NotValidErrorType exception in ctor

hikinggrass:
This fixes clangs warning: returning address of local temporary object [-Wreturn-stack-address] compiler warning

### DIFF
--- a/include/utils/error.hpp
+++ b/include/utils/error.hpp
@@ -24,14 +24,13 @@ RequestClearErrorOption string_to_request_clear_error_option(const std::string& 
 
 class NotValidErrorType : public std::exception {
 public:
-    explicit NotValidErrorType(const std::string& error_type_) :
-        error_type(error_type_), explanatory_string(fmt::format("Error type '{}' is not valid.", error_type)){};
+    explicit NotValidErrorType(const std::string& error_type) :
+        explanatory_string(fmt::format("Error type '{}' is not valid.", error_type)){};
     const char* what() const noexcept override {
         return explanatory_string.c_str();
     }
 
 private:
-    const std::string error_type;
     const std::string explanatory_string;
 };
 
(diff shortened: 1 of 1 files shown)


include/utils/error.hpp
```diff
@@ -24,13 +24,15 @@ RequestClearErrorOption string_to_request_clear_error_option(const std::string&
 
 class NotValidErrorType : public std::exception {
 public:
-    explicit NotValidErrorType(const std::string& error_type_) : error_type(error_type_){};
-    virtual const char* what() const throw() {
-        return fmt::format("Error type '{}' is not valid.", error_type).c_str();
+    explicit NotValidErrorType(const std::string& error_type_) :
+        error_type(error_type_), explanatory_string(fmt::format("Error type '{}' is not valid.", error_type)){};
```
a-w50: Is there any reason, why `error_type` should be kept as a member variable?

a-w50 (approved):

include/utils/error.hpp
hikinggrass: @andistorm ?

andistorm: `error_type` as member variable might allows to handle the exception in specific cases. Right now there is no usage of that. And it should be probably public to allow using it..

include/utils/error.hpp
```diff
@@ -24,13 +24,15 @@ RequestClearErrorOption string_to_request_clear_error_option(const std::string&
...
+    const char* what() const noexcept override {
+        return explanatory_string.c_str();
     }
 
 private:
     const std::string error_type;
```
andistorm:
```suggestion
    const std::string error_type;

private:
```

andistorm (approved):

include/utils/error.hpp
hikinggrass: I'm not so sure about making this member public, should probably be a getter. Or we remove it for now if there's no usage

a-w50: I would remove it here, if it will be needed, it can just be added

PR #130, opened 2023-11-23
Add bazel support for easier external integration

golovasteek:
Many enterprise companies use [Bazel](https://bazel.build/) to build their softwawre. Among them Qwello, [BMW](https://www.youtube.com/watch?v=Gh4SJuYUoQI), [Fernride](https://github.com/fernride/bazel-fernride-registry).

Bazel is very good suitable for large repositories, because it ensures correctness of the incremental builds. And therefor even on large code bases CI can handle testing every commit. 

To make adoption of Everest easier in such projects we may ship bazel rules for everest-framework.

In this PR bazel rules are added that allow to build everest-framework library, manager and controller.

The proposed usage is described in [bazel/README.md](https://github.com/EVerest/everest-framework/blob/9bf4dfa7a3a4ff65c0ba390fe923ba6766dd3a4b/third-party/bazel/README.md).


.github/workflows/bazel_build.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - run: echo branch name is ${{ github.ref }}
+      - name: Checkout
+        uses: actions/checkout@v4.1.0
+      - name: Mount bazel cache
+        uses: actions/cache@v3
+        with:
+          path: "~/.cache/bazel"
+          key: "bazel-cache"
```
a-w50: I think this doesn't work as you're expecting.  Once a cache has been created, it will never be overwritten again (otherwise you could get races, when multiple actions are running).  Also see https://github.com/actions/toolkit/issues/505

bazel/third-party/BUILD.curl.bazel
```diff
@@ -0,0 +1,115 @@
+# This file is copied from https://github.com/3rdparty/bazel-rules-curl/tree/f64665f4614beb19f08b45b755d7c77f487287bb
```
a-w50: I don't really understand, what is done here.  But is it using boringssl or openssl?  I guess it would make sense to keep these things consistent over the different build systems, i.e. either openssl or boringssl

.github/workflows/bazel_build.yml
```diff
@@ -0,0 +1,21 @@
+name: Bazel Build
+run-name: ${{ github.actor }} is building with bazel
+on: [push]
+jobs:
+  bazel-build:
```
SirVer, line 5: @golovasteek Github Actions default to come with bazelisk, which I think you should be using to build. See for example https://github.com/Qwello/zvt/blob/main/.github/workflows/bazel.yml, which does not do caching though.

.github/workflows/bazel_build.yml
golovasteek: At this point it's hard to guess, what would be the good cache key.

~/.cache/bazel folder doesn't grow infinitely in my experience, the obsolete stuff is getting removed from there.

What maybe makes sense to add here a platform, once/if we would build for multiple platforms here.

Do you have ideas about good cache keys for this keys?

bazel/README.md
```diff
@@ -0,0 +1,25 @@
+# Using everst-framework from bazel workspace
```
SirVer: Add something here that bazel is currently best-effort & not tested on CI, but used by Qwello for production usage. Make it clear that bazel related stuff should be pinged to  somebody at Qwello

bazel/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Add the following to your WORKSPACE file:
+```python
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+EVEREST_FRAMEWORK_REVISION = "4c0488de5eca3a9bc7c9258fc3715c5d6aab8c10"
```
SirVer: Needs a comment that this is the SHA you are pulling.

bazel/README.md
SirVer: Also, add that c++ module building is currently not implemented yet?

bazel/third-party/BUILD.curl.bazel
SirVer: Bazel is a hermetic build system, i.e. it is building the world from scratch. These are just rules to build libcurl. This is using boringssl, and we only use boringssl everywhere.

bazel/third-party/BUILD.json-schema-validator.bazel
```diff
@@ -0,0 +1,14 @@
+
```
SirVer: nit: please run buildfier on all of the build files to get rid of empty lines and indenting inconsistencies.

SirVer (approved):
Very nice, just a couple of nits.

SirVer:
Following the process that was suggested on the TSC meeting yesterday, we'd like to merge this best-effort support for Bazel which we have been using internally. Having it here makes it easier to keep in sync with future development of the framework and also makes it easy for others to use this in their own build processes. 

@corneliusclaussen @caller or any other: any inputs on this?

bazel/third-party/BUILD.curl.bazel
golovasteek: My understanding is that the "libcurl" normally comes from the system, and is not predefined. So it's up to system/distribution maintainer to choose how libcurl is build.

Here in the bazel build we have to make this choice as well. I've chose boringssl, because it seems to be more bazel-friendly (namely they have bazel support built-in).

corneliusclaussen: The main features in core for e.g. TPM based TLS for OCPP etc will use openssl3, no idea if this is supported on boringssl as well. Also google changes the API as they like, so I would not claim we officially support boringssl at all.

.github/workflows/bazel_build.yml
golovasteek: Basically `bazel` and `bazelisk` are the same in the github actions.

I see that when running bazel it downloads the right version:

```
2023/11/23 23:48:53 Downloading https://releases.bazel.build/6.4.0/release/bazel-6.4.0-linux-x86_64...
```

corneliusclaussen:
> @corneliusclaussen @caller or any other: any inputs on this?

I'm good with this, but I also think we should put a warning in the Readme or so stating that this build method is wip and your milage may vary.

corneliusclaussen:
To get this merged pls look at the comments and especially at the caching key and mark as "Ready for review" when you are done.

.github/workflows/bazel_build.yml
golovasteek: Finally I got your point.

Indeed it works not as I was expecting.

Fixed from exampes: https://github.com/actions/cache/blob/main/examples.md#---bazel

bazel/README.md
golovasteek: No modules are supported in this PR

bazel/third-party/BUILD.curl.bazel
golovasteek: I've added warning about BoringSSL to Readme.

We need more details regarding the requirements. I'm up to support openssl1 or 3 eventually.

bazel/third-party/BUILD.json-schema-validator.bazel
golovasteek: Done

golovasteek:
I've addressed all comments.

corneliusclaussen (approved):

PR #137, opened 2023-12-18
fix(save_config): non existing strings

DerMolly:
If the string is an empty string we don't strip the quotes and make the resulting saved yaml sytactically wrong. Unfortunately this leaves the quotes around the key as well. This should be investigated further at a later point.

This fixes the problem descriped in https://lists.lfenergy.org/g/everest/message/1025


CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ find_package(everest-cmake 0.1 REQUIRED
 )
 
 # options
-option(BUILD_TESTING "Run unit tests" OFF)
+option(FRAMEWORK_TESTING "Enable unit tests" OFF)
```
gberardi-pillar: Why are we changing the variable that CTest relies on? 

`BUILD_TESTING` is the variable used by CTest, and so I presume that we want to keep that standardized: https://cmake.org/cmake/help/latest/module/CTest.html

tests/CMakeLists.txt
```diff
@@ -1,15 +1,25 @@
 include(Catch)
 
-add_executable(tests test_config.cpp)
+set (TEST_PREFIX "test_${PROJECT_NAME}")
```
gberardi-pillar: Does this change need to be part of this review? I am specifically addressing the test name uniqueness in https://github.com/EVerest/everest-framework/pull/140

DerMolly: I will remove this.

CMakeLists.txt
DerMolly: The idea was to make it possible to just test specific parts of the project and not the whole, but I guess this needs to addressed in a more general way. I remove it for now…

src/controller/command_api.cpp
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2020 - 2022 Pionix GmbH and Contributors to EVerest
+// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
```
gberardi-pillar: Should this be 2024 now?

tests/CMakeLists.txt
```diff
@@ -11,14 +11,22 @@ target_sources(tests PRIVATE
...
-target_link_libraries(tests
+add_subdirectory(controller)
+
+add_executable(
+    tests
+        test_config.cpp
+)
+
+
+target_link_libraries(${CONFIG_TEST}
```
gberardi-pillar: If you are reverting, ${CONFIG_TEST} should be tests here and on line 29.

hikinggrass (approved):

PR #138, opened 2023-12-19
fix: Add node21 as potential include path

nzbr:
Adds `node21` to the candidates for `NODEJS_INCLUDE_DIR`, so that cmake is able to find the node headers on OpenSUSE-Tumbleweed


hikinggrass:
Thx, you're missing the signoff in your commit, that why the DCO check complains. Just add that to your commit and rebase to current main and we can merge this :)

hikinggrass (approved):

a-w50 (approved):

PR #139, opened 2023-12-19
Use poll for mqtt io instead of sleeping

corneliusclaussen:


corneliusclaussen:
One bug has been found that needs to be fixed before merging. After a while mqttc gives socket errors

corneliusclaussen:
> One bug has been found that needs to be fixed before merging. After a while mqttc gives socket errors

fixed

hikinggrass (approved):

PR #140, opened 2023-12-19
Name test target after repo to make it unique.

gberardi-pillar:
everest-core cannot build with tests (BUILD_TESTING=ON) due to submodules having generic names that collide when CMake runs.

This change makes the test target for everest-framework unique by naming it after the repo,  a pattern that can be followed for all repos.

I also used the Modern CMake pattern of enabling tests only if the project is the main project (not a child of a parent project) or if it is explicitly specified with cmake -Deverest-framework_BUILD_TESTING=ON to allow flexibility.

NOTE: This change does not enable automated builds and tests, as there are other PRs to handle that work.


gberardi-pillar:
@shankari Here's a minor change to review.

tests/CMakeLists.txt
```diff
@@ -1,24 +1,25 @@
 include(Catch)
 
-add_executable(tests)
+set(TEST_TARGET_NAME ${PROJECT_NAME}_tests)
+add_executable(${PROJECT_NAME})
```
Dominik-K: Shouldn't it be 
```suggestion
add_executable(${TEST_TARGET_NAME})
```
?

Dominik-K (commented):
Hi @gberardi-pillar, thanks for your contribution.

IMO, it's a good idea to _prefix_ the `tests`-names. 

However, I think the root-cause is that `BUILD_TESTING` is used in all dependencies, too. So when you activate it here (in `everest-framework`), then it's activated also in many/all(?) dependencies. And then the clash happens. 

Is it intended that `BUILD_TESTING` is used generic in all repos? @hikinggrass @andistorm @a-w50 (Note for myself: If so, revert it [in this PR](https://github.com/EVerest/libocpp/pull/301/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR106))

tests/CMakeLists.txt
gberardi-pillar: Yes, good catch. I don't know how I missed that in my latest typo fix push.

gberardi-pillar:
> Hi @gberardi-pillar, thanks for your contribution.
> 
> IMO, it's a good idea to _prefix_ the `tests`-names.

To be clear, do you mean change the test target name from `${PROJECT_NAME}_tests` to `tests_${PROJECT_NAME}`? If that is the norm, I can change it.
 
> However, I think the root-cause is that `BUILD_TESTING` is used in all dependencies, too. So when you activate it here (in `everest-framework`), then it's activated also in many/all(?) dependencies. And then the clash happens.
> 
> Is it intended that `BUILD_TESTING` is used generic in all repos? @hikinggrass @andistorm @a-w50 (Note for myself: If so, revert it [in this PR](https://github.com/EVerest/libocpp/pull/301/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR106))

`BUILD_TESTING` is the variable used by CTest, and so I presume that we want to keep that standardized: https://cmake.org/cmake/help/latest/module/CTest.html 

Now, there might be some discussion about how to handle testing overall. For instance, if I build everest-framework and run its tests, should doing so also run the tests of its dependencies, which makes sense since I am also building those dependencies, or should I trust that those dependencies ran their own tests before they reached the point of being used by a parent repo and have some way of indicating that the unit tests I want to run are exclusively the current project in that repo?

But for now, presuming that we are building all of the code, as C++ generally wants all code modules to be compiled with the same flags, I think it makes sense to run all of the tests with it, and setting `BUILD_TESTING=ON` in a parent module should run the tests in all submodules as well.

a-w50:
Hi,

this PR is quite similar to  #137 , maybe we can use that one instead of this one.

gberardi-pillar:
> Hi,
> 
> this PR is quite similar to #137 , maybe we can use that one instead of this one.

That PR is a draft (so I don't know how long it will be before it is in a state to actually be reviewed), and it is supposed to be fixing an issue with a config variable being null (and so seems to be doing too much). This PR is more specific to changing the test target name.

CMakeLists.txt
```diff
@@ -145,7 +145,7 @@ endif ()
 
 # testing
 # FIXME (aw): move testing logic into tests/CMakeLists.txt
-if(BUILD_TESTING)
+if((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME OR ${PROJECT_NAME}_BUILD_TESTING) AND BUILD_TESTING)
```
Dominik-K, line 156:
```suggestion
if( (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING)  OR ${PROJECT_NAME}_BUILD_TESTING )
```

Either `BUILD_TESTING` is set globally (therefore filter it) or `${PROJECT_NAME}_BUILD_TESTING` is used.

Dominik-K (dismissed):

Dominik-K:
@gberardi-pillar 
> > IMO, it's a good idea to _prefix_ the `tests`-names.
> 
> To be clear, do you mean change the test target name from `${PROJECT_NAME}_tests` to `tests_${PROJECT_NAME}`? If that is the norm, I can change it.

No, just like you did, i.e.  `PROJECT_NAME_tests`. I just wanted to support that approach :+1:  Sorry for the confusion.

With prefixes you can see at first step which file belongs to which project. I don't see a use-case where a sorting by `tests_...` is needed. But perhaps someone else more used to the CMake stuff. E.g. @a-w50 @hikinggrass 

[You mentioned](https://lists.lfenergy.org/g/everest/message/1167) (in the mailing-list) https://cliutils.gitlab.io/modern-cmake/chapters/testing.html. That looks like a good document for best-practices. In there, it also uses the prefix-approach.

So I would say it's good to go :+1:

Dominik-K:
Let's discuss in general at
- https://github.com/EVerest/EVerest/issues/129

shankari:
I see that there is a longer-term issue, but it is generally in line with this approach and is primarily tracking its application to multiple repos. So let's get this merged and get the ball rolling on additional changes!

I do note that the "Bazel Build" workflow is currently broken although the failure does not seem to be related to this change. 
@golovasteek you last committed the bazel build CI; is there a plan to fix the action?

shankari (approved):

golovasteek:
> I do note that the "Bazel Build" workflow is currently broken although the failure does not seem to be related to this change.
> @golovasteek you last committed the bazel build CI; is there a plan to fix the action?

I've fixed the bazel-build yesterday, it should work after rebase onto main.

CMakeLists.txt
```diff
@@ -151,7 +151,7 @@ endif ()
...
```
andistorm, line 156:
```suggestion
if((${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME} OR ${PROJECT_NAME}_BUILD_TESTING) AND BUILD_TESTING)
```

CMakeLists.txt
andistorm, line 1: Please update target names for `gcovr` and `lcovr`
```
    setup_target_for_coverage_gcovr_html(
        NAME ${PROJECT_NAME}_gcovr_coverage
```
```
    setup_target_for_coverage_lcov(
        NAME ${PROJECT_NAME}_lcov_coverage
```

andistorm, line 1: please add the options:
```
option(${PROJECT_NAME}_BUILD_TESTING "Build unit tests, used if included as dependency" ON)
option(BUILD_TESTING "Build unit tests, used if standalone project" ON)
```

andistorm (changes requested):

andistorm (approved):

PR #151, opened 2024-01-26
Feature/add watchdog helper

corneliusclaussen:
Adds watchdog functionality to the framework

Each module has a WatchdogSupervisor in its main module class that sends out watchdog feed heartbeats via MQTT.
This can be monitored by the manager (not yet implemented here). This at least ensures that the watchdog supervising thread in the module is still alive and MQTT communication works.

The module code can register multiple internal watchdogs, which is useful if it has looping threads.


hikinggrass:
According to https://github.com/EVerest/everest-core/pull/514 there's no ongoing work atm

PR #158, opened 2024-02-20
Link stdc++fs for GCC<9

### LINKED ISSUE #157: Support GCC<9
GCC < 9 needs std++fs if std::filesystem is used, we should link that in that case

hikinggrass:
Fixes https://github.com/EVerest/everest-framework/issues/157

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,13 @@ pkg_check_modules(libcap
     libcap
 )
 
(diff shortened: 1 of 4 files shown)


a-w50 (approved):
Looks good, still would sort the link libraries from less specific to more specific like with the  `#include` directives - first c/c++ standard libraries, system libraries, tool libraries and user libraries ...

PR #161, opened 2024-02-21
MQTT event loop optimization and added support for Unix Domain Sockets for MQTT communication

florinmihut:
* Added support for Unix Domain Sockets for MQTT communication
   - Created a new config option 'mqtt_broker_socket_path' where one can provide the socket path
   - The config option mutual exclusive to the normal TCP socket and will raise an exception if both are provided in the config file (or as env variable)
   - There is no default value for the UDS path, it must be provided in the config file
   - We keep the backward compatibility, thus if nothing is provided in the config file, the default MQTT values are used (localhost:1883)
   - Beside functionality added, this provides an optimization over the TCP connection (observed on slow systems around 5%)
   - Updated the config schema to for the new option
* Optimization made in the MQTT event loop (we moved from executing the event loop from 100ms to 300s)
   - Normally the keep alive message is send by the MQTT lib, in certain cases however this might not work unless you call mqtt_sync often and on regular basis (few ms)
   - We decreased the event loop execution frequency drastically (on slow systems we can see a decrease of CPU usage around 10%)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14)
 
 project(everest-framework
-    VERSION 0.10
+    VERSION 0.10.1
     DESCRIPTION "The open operating system for e-mobility charging stations"
 	LANGUAGES CXX C
 )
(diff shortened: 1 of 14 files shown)


florinmihut:
Looks to me that Codacy has some false positives, I tried to "fix" them to make it see that is a false positive but in vain.

hikinggrass (approved):

PR #170, opened 2024-03-28
Improve disconnect procedure of MQTTAbstraction

andistorm:
* poll() function returns now on disconnect

closes https://github.com/EVerest/everest-utils/issues/125

### DIFF
--- a/lib/mqtt_abstraction_impl.cpp
+++ b/lib/mqtt_abstraction_impl.cpp
@@ -220,6 +220,7 @@ std::future<void> MQTTAbstractionImpl::spawn_main_loop_thread() {
                 if (retval >= 0) {
                     // data available to send (the notifier writes, we should be ready to read)
                     if (retval > 0) {
+                        // check for disconnect event
                         if (pollfds[2].revents & POLLIN) {
                             break;
                         }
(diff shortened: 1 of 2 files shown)


lib/mqtt_abstraction_impl.cpp
```diff
@@ -200,12 +211,18 @@ std::future<void> MQTTAbstractionImpl::spawn_main_loop_thread() {
...
+                const int nfds = 3;
+                struct pollfd pollfds[nfds] = {{this->mqtt_socket_fd, POLLIN, 0},
+                                               {this->event_fd, POLLIN, 0},
+                                               {this->disconnect_event_fd, POLLIN, 0}};
+                auto retval = ::poll(pollfds, nfds, mqtt_poll_timeout_ms);
 
                 if (retval >= 0) {
                     // data available to send (the notifier writes, we should be ready to read)
                     if (retval > 0) {
+                        if (pollfds[2].revents & POLLIN) {
```
hikinggrass, line 224: Can you add a comment that explains that we're disconnecting here?

hikinggrass (changes requested):

lib/mqtt_abstraction_impl.cpp
andistorm: Done

hikinggrass (approved):

PR #171, opened 2024-04-02
Rust: Link against gcov

dorezyuk:
When we build with BUILD_TESTING, we're adding the flags for gcov and need to subsequently link against gcov. This commit fixes the behavior for rust


dorezyuk:
requires https://github.com/EVerest/everest-framework/pull/172

dorezyuk:
@hikinggrass Can you review this?

hikinggrass (approved):

PR #172, opened 2024-04-02
Bazel: update boost depenedency

dorezyuk:
In the wake of the xz backdoor update the bazel-boost dependency to not depend on the tukaani github project


hikinggrass:
Fixed in https://github.com/EVerest/everest-framework/pull/173

Issue #174, opened 2024-04-03
Support Multislot Connection in Rust


golovasteek:
At the moment only one connection is exposed to Rust model for the case of multi-slot connection.

Need to re-design rust wrappers and rust codegen, to  support multi-slot connections.

SirVer:
I know that @dorezyuk has started working on this. I cannot assign him though

hikinggrass:
Implemented in #176

PR #176, opened 2024-04-03
Rust: add slots support

dorezyuk:
Adds Rust support for multiple slots.

## Summary
The proposed interface is
* if `max_connections` or `min_connections` is not defined we're defaulting for every value to 1
* if `max_connections` and `min_connections` are evaluated to 1 nothing changes to the current interface
* if `max_connections` and `min_connections` are not one, we define a vec/array (depending if the size is known at compile time) of `ClientPublishers`. The used slot is exposed to the user though the `context` - and use the same instance of the `ClientSubscriber` for the entire slots

## Examples
```yaml
# in the manifest.yaml
# ...
requires
  example_interface:
    interface: example
    min_connections: 0
    max_connections: 10
```

```rs
// In the main.rs
include!(concat!(env!("OUT_DIR"), "/generated.rs"));
use generated::{ExampleClientSubscriber, Context};

struct MyModule;

impl ExampleClientSubscriber for MyModule {
    fn on_max_current(&self, context: &Context, value: f64) {
       println!("Received a connection from the slot {}.. replying to it by using something", context.index);
       context.publisher.example_interface_slots[context.index].uses_something(format!("value {value}"));
    }
}

```
For disambiguating the slots at the `ClientSubscriber` we change the interface for multiple slots (if these is only one stot, nothing changes) for the `Module` to be a `FnMut` similar to the [`core::array::from_fn`](https://doc.rust-lang.org/stable/core/array/fn.from_fn.html) in the Rusts standard library. It allows the user to decide if they want to use different instances of `ClientPublisher` or the same for every slot.

```rs
struct MyModuleWithSomeData{
   index: usize,
};

impl ExampleClientSubscriber for MyModuleWithSomeData {
    fn on_max_current(&self, context: &Context, value: f64) {
         assert_eq!(self.index, context.index);
    }
}

// ...

fn main() {
      let my_module = Arc::new(MyModule);
      let _module = Module::new(
            |_index: usize| {
              // Below we just return the same instance for every slot.
              // We could also return Arc::new(MyModuleWithSomeData{index: _index}) here if
              // we want every slot to be handled by a dedicated instance
               my_module.clone() 
            }
      );
}
```

Addresses https://github.com/EVerest/everest-framework/issues/174


dorezyuk:
@SirVer can you check it

SirVer:
Will do early this week.

SirVer (approved):
Looks good to me, nice solution I think.

hikinggrass (approved):

PR #177, opened 2024-04-04
Add version information

hikinggrass:
This outputs version information in the manager on startup and when passed a --version parameter on the command line


andistorm (approved):
LGTM

PR #189, opened 2024-06-04
Use check_code() to make sure that cmds are implemented in everestpy

hikinggrass:
This helps prevent that modules calling into everestpy modules "hang" when calling non-implemented cmds


andistorm (approved):
Tested :+1:

PR #190, opened 2024-06-05
Log raised errors

andistorm:
* Bump version to v0.14.1
* Log when error is raised/cleared
* Log warning and do nothing when error is already active/cleared instead of throwing exception

### DIFF
--- a/lib/error/error_manager_impl.cpp
+++ b/lib/error/error_manager_impl.cpp
@@ -8,6 +8,8 @@
 #include <utils/error/error_exceptions.hpp>
 #include <utils/error/error_type_map.hpp>
 
+#include <everest/logging.hpp>
+
 #include <list>
 #include <memory>
 
@@ -43,6 +45,7 @@ void ErrorManagerImpl::raise_error(const Error& error) {
     }
     database->add_error(std::make_shared<Error>(error));
     this->publish_raised_error(error);
+    EVLOG_error << "Error raised, type: " << error.type << ", sub_type: " << error.sub_type << ", message: " << error.message;
 }
 
 std::list<ErrorPtr> ErrorManagerImpl::clear_error(const ErrorType& type, const bool clear_all) {
--- a/lib/error/error_manager_impl.cpp
+++ b/lib/error/error_manager_impl.cpp
@@ -41,7 +41,8 @@ void ErrorManagerImpl::raise_error(const Error& error) {
         }
     }
     if (!can_be_raised(error.type, error.sub_type)) {
-        throw EverestArgumentError("Error type " + error.type + " is already active.");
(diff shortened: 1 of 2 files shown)


corneliusclaussen (changes requested):
As far as I can see that's only for raising. We should log the clearing as well

andistorm:
> As far as I can see that's only for raising. We should log the clearing as well

I added this and I also turned exception throwing into logged warnings

lib/error/error_manager_impl.cpp
```diff
@@ -51,19 +56,25 @@ std::list<ErrorPtr> ErrorManagerImpl::clear_error(const ErrorType& type, const b
...
+        ss << "  - " << error->sub_type << std::endl;
     }
+    EVLOG_info << ss.str();
     return res;
 }
 
 std::list<ErrorPtr> ErrorManagerImpl::clear_error(const ErrorType& type, const ErrorSubType& sub_type) {
     if (!can_be_cleared(type, sub_type)) {
-        throw EverestArgumentError("Error can't be cleared, because it is not active.");
+        EVLOG_warning << "Error can't be cleared, because it is not active.";
```
barsnick: Probably also "Error**s** can't", as in the other implementation of `clear_error()`.

This should also print type and subtype:

```suggestion
        EVLOG_warning << "Errors can't be cleared, because type " << type << ", sub_type " << sub_type << " is not active.";
```

lib/error/error_manager_impl.cpp
```diff
@@ -39,10 +41,13 @@ void ErrorManagerImpl::raise_error(const Error& error) {
         }
     }
     if (!can_be_raised(error.type, error.sub_type)) {
-        throw EverestArgumentError("Error type " + error.type + " is already active.");
+        EVLOG_warning << "Error can't be raised, because it is already active.";
```
a-w50: Probably still log, which error type / subtype

a-w50: Ah, just saw that @barsnick mentioned this already below.

lib/error/error_manager_impl.cpp
andistorm: > Probably also "Errors can't", as in the other implementation of clear_error().

In this case it is actually a function to exactly clear one error, the other functions can clear multiple ones

> This should also print type and subtype:

Good point, thanks! I added it

corneliusclaussen:
as far as I can see there are three throws left that should be replaced

corneliusclaussen:
> as far as I can see there are three throws left that should be replaced

We can fix this in a separate PR

corneliusclaussen (approved):

PR #193, opened 2024-06-07
Ignore unknown config options instead of exiting

corneliusclaussen:
Unknown config options prevented the start up of EVerest. This can easily happen during updates of everest if a config option was removed but the config file was not adapted. As this can also happen in a user config overlay, it should not prevent the startup. Logging an error message should be enough.

### DIFF
--- a/lib/config.cpp
+++ b/lib/config.cpp
@@ -21,7 +21,6 @@ namespace Everest {
 class ConfigParseException : public std::exception {
 public:
     enum ParseErrorType {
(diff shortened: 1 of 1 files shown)


hikinggrass:
@corneliusclaussen I've changed this slightly to be able to log a more descriptive error like before, just without the exception

corneliusclaussen:
Looks good, just tested it again

hikinggrass (approved):

PR #194, opened 2024-06-07
Remove some more throws in error framework

corneliusclaussen:
Remove some more throws in error framework. They should never really happen, but if it does it is not critical enough to shut down the whole system.


lib/error/error_database_map.cpp
```diff
@@ -60,7 +60,7 @@ std::list<ErrorPtr> ErrorDatabaseMap::get_errors_no_mutex(const std::list<ErrorF
                     return error->severity < Severity::High;
                 } break;
                 }
-                throw std::out_of_range("No known condition for provided enum of type SeverityFilter.");
+                EVLOG_error << "No known condition for provided enum of type SeverityFilter.";
```
andistorm, line 62:
```suggestion
                EVLOG_error << "No known condition for provided enum of type SeverityFilter.";
                return false;
```

lib/error/error_manager_impl.cpp
```diff
@@ -34,10 +34,12 @@ void ErrorManagerImpl::raise_error(const Error& error) {
     if (validate_error_types) {
         if (std::find(allowed_error_types.begin(), allowed_error_types.end(), error.type) ==
             allowed_error_types.end()) {
-            throw EverestArgumentError("Error type " + error.type + " is not allowed to be raised.");
+            EVLOG_error << "Error type " << error.type << " is not allowed to be raised. Ignoring.";
+            return;
```
andistorm, line 44: Not sure if we should do not crash here, since the caller isn
t aware of that error isn't raised. In the following the error can't be handled.  Which might lead to critical undefined behaviour..

lib/error/error_manager_impl.cpp
```diff
@@ -34,10 +34,12 @@ void ErrorManagerImpl::raise_error(const Error& error) {
...
         }
         if (!this->error_type_map->has(error.type)) {
-            throw EverestArgumentError("Error type " + error.type + " is not known.");
+            EVLOG_error << "Error type " << error.type << " is not known.";
+            return;
```
andistorm: This if statement doesn't really make sense at all, it should be removed, since an error_type which is in `allowed_error_types` should be in ` error_type_map`, too.

As a replacement for this check it would make sense to validate all types in ` allowed_error_types` once in constructor

lib/error/error_manager_impl.cpp
```diff
@@ -81,7 +83,8 @@ std::list<ErrorPtr> ErrorManagerImpl::clear_error(const ErrorType& type, const E
     std::list<ErrorFilter> filters = {ErrorFilter(TypeFilter(type)), ErrorFilter(SubTypeFilter(sub_type))};
     std::list<ErrorPtr> res = database->remove_errors(filters);
     if (res.size() > 1) {
-        throw EverestBaseLogicError("There are more than one matching error, this is not valid");
+        EVLOG_error << "There are more than one matching error, this is not valid";
+        return {};
```
andistorm, line 89: To ignore this case it should be checked with `database->get_errors(filters)` first, to then don't remove .

Anyways this means that the database is in a corrupted state which is not recoverable, so it should crash/throw

lib/error/error_manager_req.cpp
```diff
@@ -50,11 +50,13 @@ void ErrorManagerReq::subscribe_all_errors(const ErrorCallback& callback, const
 
 void ErrorManagerReq::on_error_raised(const Error& error) {
     if (std::find(allowed_error_types.begin(), allowed_error_types.end(), error.type) == allowed_error_types.end()) {
-        throw EverestBaseLogicError("Error can't be raised by module_id " + error.origin.module_id +
-                                    " and implemenetation_id " + error.origin.implementation_id);
+        EVLOG_error << "Error can't be raised by module_id " << error.origin.module_id << " and implemenetation_id "
+                    << error.origin.implementation_id << ", ignored.";
+        return;
     }
```
andistorm, line 64: This check kind of duplicates the check on `ErrorManagerImpl`-side.
In case of passing the first check this means to be in a very corrupted state, which should crash/throw. Since a raised error is not accepted and not handled.

lib/error/error_manager_req.cpp
```diff
@@ -50,11 +50,13 @@ void ErrorManagerReq::subscribe_all_errors(const ErrorCallback& callback, const
...
     if (!error_type_map->has(error.type)) {
-        throw EverestBaseLogicError("Error type '" + error.type + "' is not defined");
+        EVLOG_error << "Error type '" << error.type << "' is not defined, ignored.";
+        return;
     }
```
andistorm, line 68: same as above

lib/error/error_manager_req.cpp
```diff
@@ -64,11 +66,13 @@ void ErrorManagerReq::on_error_cleared(const Error& error) {
     std::list<ErrorFilter> filters = {ErrorFilter(TypeFilter(error.type)), ErrorFilter(SubTypeFilter(error.sub_type))};
     std::list<ErrorPtr> res = database->remove_errors(filters);
     if (res.size() < 1) {
-        throw EverestBaseLogicError("Error wasn't raised, type: " + error.type + ", sub_type: " + error.sub_type);
+        EVLOG_error << "Error wasn't raised, type: " << error.type << ", sub_type: " << error.sub_type << ", ignored.";
+        return;
     }
```
andistorm, line 79: This means to be in a corrupted/unrecoverable state

lib/error/error_manager_req.cpp
```diff
@@ -64,11 +66,13 @@ void ErrorManagerReq::on_error_cleared(const Error& error) {
...
     if (res.size() > 1) {
-        throw EverestBaseLogicError("More than one error is cleared, type: " + error.type +
-                                    ", sub_type: " + error.sub_type);
+        EVLOG_error << "More than one error is cleared, type: " << error.type << ", sub_type: " << error.sub_type
+                    << ", ignored.";
+        return;
     }
```
andistorm, line 84: This means to be in a corrupted/unrecoverable state

andistorm (changes requested):

lib/error/error.cpp
```diff
@@ -75,7 +76,8 @@ std::string severity_to_string(const Severity& s) {
     case Severity::High:
         return "High";
     }
-    throw std::out_of_range("No known string conversion for provided enum of type Severity.");
+    EVLOG_error << "No known string conversion for provided enum of type Severity:" << s << ". Defaulting to High.";
```
andistorm: `Severity s` is not a string
```suggestion
    EVLOG_error << "No known string conversion for provided enum of type Severity. Defaulting to High.";
```

.. same for the following enums

lib/error/error_manager_impl.cpp
corneliusclaussen: The problem is that it is even possible at compile time to do this. Simply crashing is not well defined behaviour either. So to properly fix this it needs to be detected at compile time. Until then, in my mind ignoring (and logging this in the log) it still better then tearing down the whole system with an uncaught exception.

lib/error/error_manager_impl.cpp
corneliusclaussen: Sounds good, can you change this?

lib/error/error_manager_impl.cpp
corneliusclaussen: It should not, see above. The problem is that a crash does not solve the problem, after restart, it is still there and will crash again, ending up in a restart loop. If it can be guaranteed that it never happens, it does not need to be checked here. Apparently it can happen, and then it needs to be dealt with somehow, e.g. removing one of them in the database.

lib/error/error_manager_req.cpp
corneliusclaussen: See above, how does a crash fix this scenario?

andistorm (approved):

hikinggrass (approved):

PR #198, opened 2024-07-24
Add 3 tier model mapping support

hikinggrass:
Add module level evse and connector mappings

Add 3 tier model mappings to errors

### DIFF
--- a/include/utils/types.hpp
+++ b/include/utils/types.hpp
@@ -84,9 +84,10 @@ struct TelemetryConfig {
     int id;
 };
 
+/// \brief A Mapping that can be used to map a module or implementation to a specific EVSE or optionally to a Connector
 struct Mapping {
-    int evse;
-    std::optional<int> connector;
+    int evse;                     ///< The EVSE id
+    std::optional<int> connector; ///< An optional Connector id
 
     Mapping(int evse) : evse(evse) {
     }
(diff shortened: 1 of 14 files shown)


schemas/config.yaml
```diff
@@ -147,6 +147,36 @@ properties:
             default: {}
             # don't allow arbitrary additional properties
             additionalProperties: false
+          evse:
```
Pietfried, line 150: nit / personal preference: `evse_id`

schemas/config.yaml
```diff
@@ -147,6 +147,36 @@ properties:
...
+            description: evse this module id maps to
+            type: integer
+          connector:
```
Pietfried, line 153: nit / personal preference: `connector_id`

include/utils/types.hpp
```diff
@@ -84,6 +84,22 @@ struct TelemetryConfig {
     int id;
 };
 
+struct Mapping {
```
Pietfried, line 88: I think a brief description for the struct would help here, same for `ModuleTierMappings`

Pietfried (approved):
Added a few suggestions

Pietfried:
We should also add this new functionality to the documentation, maybe here: https://everest.github.io/nightly/general/05_existing_modules.html#everest-module-configurations 
@krealyt

schemas/config.yaml
hikinggrass: my thinking here was that this gets a bit verbose, but I don't mind one way or the other. But if we change it we should also change it in the impl/provides mapping

include/utils/types.hpp
hikinggrass: I'll add some more documentation for everything that was added in this pr :+1:

PR #201, opened 2024-08-19
Config restructure: distribute module configs via MQTT

hikinggrass:
Changes to reduce configuration parsing effort by parsing it once by the manager and then distribute it to the modules via MQTT.

This is achieved by first parsing the MQTT settings from the config, passing these to the modules whilst spawning them. The modules themselves then ask for their module config via MQTT, which is in turn provided to them from the manager.

schemas, manifests, types, interfaces, error definitions are published as retained topics that modules can access when needed.


CMakeLists.txt
```diff
@@ -175,18 +182,21 @@ if((${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME} OR ${PROJECT_NAME}_BUILD_TEST
...
 
+    set(GCOVR_ADDITIONAL_ARGS "${PROJECT_SOURCE_DIR}" "--gcov-ignore-errors=all")
     setup_target_for_coverage_gcovr_html(
         NAME ${PROJECT_NAME}_gcovr_coverage
         EXECUTABLE ctest --output-on-failure
-        EXCLUDE "${CMAKE_SOURCE_DIR}/tests" "${CMAKE_BINARY_DIR}"
-        DEPENDENCIES tests everest
+        BASE_DIRECTORY "${PROJECT_SOURCE_DIR}"
+        DEPENDENCIES ${PROJECT_NAME}_tests everest::framework
+        EXCLUDE "${CMAKE_SOURCE_DIR}/tests" "${CMAKE_BINARY_DIR}/*" "${CPM_SOURCE_CACHE}"
```
a-w50: Why is the tests folder excluded?

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+    const auto module_id = std::getenv("EV_MODULE");
+    if (module_id == nullptr) {
+        throw std::runtime_error("EV_MODULE needed for everestpy");
+    }
+
+    return module_id;
 }
 
-RuntimeSession::RuntimeSession() : RuntimeSession(get_ev_prefix_from_env(), get_ev_conf_file_from_env()) {
+static std::shared_ptr<Everest::MQTTSettings> get_mqtt_settings_from_env() {
```
a-w50: this function should return by value, there is no need here (yet) to create a shared pointer

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+        throw std::runtime_error("EV_LOG_CONF_FILE needed for everestpy");
     }
 
-    return config_file;
+    return logging_config_file;
 }
 
-RuntimeSession::RuntimeSession(const std::string& prefix, const std::string& config_file) :
-    rs(std::make_shared<Everest::RuntimeSettings>(prefix, config_file)), config(create_config_instance(rs)) {
+static std::string get_ev_module_from_env() {
```
a-w50: `get_ev_module_from_env` and `get_ev_log_conf_file_from_env` do the same.  I would suggest to write a common function, which gets the required env string per parameter

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+    if (module_id == nullptr) {
+        throw std::runtime_error("EV_MODULE needed for everestpy");
+    }
+
+    return module_id;
 }
 
-RuntimeSession::RuntimeSession() : RuntimeSession(get_ev_prefix_from_env(), get_ev_conf_file_from_env()) {
+static std::shared_ptr<Everest::MQTTSettings> get_mqtt_settings_from_env() {
+    const auto mqtt_everest_prefix = std::getenv("EV_MQTT_EVEREST_PREFIX");
```
a-w50: I would move the hardcoded string literals like `EV_MQTT_EVEREST...`  to a common header as constants for better visibility.  Probably they should be also used in everestjs etc.

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+    const auto mqtt_external_prefix = std::getenv("EV_MQTT_EXTERNAL_PREFIX");
+    const auto mqtt_broker_socket_path = std::getenv("EV_MQTT_BROKER_SOCKET_PATH");
+    const auto mqtt_broker_host = std::getenv("EV_MQTT_BROKER_HOST");
+    const auto mqtt_broker_port = std::getenv("EV_MQTT_BROKER_PORT");
```
a-w50: For better usability I probably would define only a single environment variable like:
`EV_MQTT_URI`, which could look like `socket://path_to_socket:0/prefix` or `tcp://localhost:1883/prefix` or something similar.  That way, tests don't need to set and know about the different env variables

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+
+    std::shared_ptr<Everest::MQTTSettings> mqtt_settings;
+    if (mqtt_broker_socket_path == nullptr) {
+        if (mqtt_broker_host == nullptr or mqtt_broker_port == nullptr) {
+            throw std::runtime_error("If EV_MQTT_BROKER_SOCKET_PATH is not set EV_MQTT_BROKER_HOST and "
```
a-w50, line 40: The error message is really precise.

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+                                     "EV_MQTT_BROKER_PORT are needed for everestpy");
+        }
+        auto mqtt_broker_port_ = Everest::defaults::MQTT_BROKER_PORT;
+        try {
+            mqtt_broker_port_ = std::stoi(mqtt_broker_port);
+        } catch (...) {
+            EVLOG_warning << "Could not parse MQTT broker port, using default: " << mqtt_broker_port_;
+        }
+        mqtt_settings = std::make_shared<Everest::MQTTSettings>(mqtt_broker_host, mqtt_broker_port_,
+                                                                mqtt_everest_prefix, mqtt_external_prefix);
```
a-w50: `mqtt_external_prefix` is not checked, would probably lead to undefined behaviour

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+            throw std::runtime_error("If EV_MQTT_BROKER_SOCKET_PATH is not set EV_MQTT_BROKER_HOST and "
+                                     "EV_MQTT_BROKER_PORT are needed for everestpy");
+        }
+        auto mqtt_broker_port_ = Everest::defaults::MQTT_BROKER_PORT;
+        try {
+            mqtt_broker_port_ = std::stoi(mqtt_broker_port);
+        } catch (...) {
+            EVLOG_warning << "Could not parse MQTT broker port, using default: " << mqtt_broker_port_;
+        }
+        mqtt_settings = std::make_shared<Everest::MQTTSettings>(mqtt_broker_host, mqtt_broker_port_,
```
a-w50: instead of declaring `mqtt_settings` beforehand, it could be just returned here directly

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+    auto module_id = get_ev_module_from_env();
+
+    namespace fs = std::filesystem;
+    fs::path logging_config_file = Everest::assert_file(get_ev_log_conf_file_from_env(), "Default logging config");
+    Everest::Logging::init(logging_config_file.string(), module_id);
+
+    this->mqtt_settings = get_mqtt_settings_from_env();
+
+    auto result = Everest::ModuleConfig::get_config(mqtt_settings, module_id);
+    this->rs = std::make_shared<Everest::RuntimeSettings>(result.at("settings"));
```
a-w50: no need for shared pointer!

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
-    Everest::Logging::init(rs->logging_config_file);
-    return std::make_unique<Everest::Config>(rs);
+/// This is just kept for compatibility
+RuntimeSession::RuntimeSession(const std::string& prefix, const std::string& config_file) : RuntimeSession() {
+    EVLOG_warning << "everestpy: Usage of the old RuntimeSession ctor detected, config is now loaded via MQTT not via "
+                     "the provided config_file.";
+}
+
+RuntimeSession::RuntimeSession() {
+    auto module_id = get_ev_module_from_env();
```
a-w50: const

everestpy/src/everest/misc.cpp
```diff
@@ -14,26 +14,71 @@ static std::string get_ev_prefix_from_env() {
...
+
+    namespace fs = std::filesystem;
+    fs::path logging_config_file = Everest::assert_file(get_ev_log_conf_file_from_env(), "Default logging config");
+    Everest::Logging::init(logging_config_file.string(), module_id);
+
+    this->mqtt_settings = get_mqtt_settings_from_env();
+
+    auto result = Everest::ModuleConfig::get_config(mqtt_settings, module_id);
```
a-w50: const

everestpy/src/everest/misc.hpp
```diff
@@ -17,15 +17,18 @@ class RuntimeSession {
         return rs;
     }
 
+    std::shared_ptr<Everest::MQTTSettings> get_mqtt_settings() const {
```
a-w50: no shared pointer! Just return const ref.
Also note that in this case function signature is misleading: it is declared as `const`, but it returns a non-const shared pointer to its member!

everestpy/src/everest/misc.hpp
```diff
@@ -17,15 +17,18 @@ class RuntimeSession {
...
+        return mqtt_settings;
+    }
+
     Everest::Config& get_config() const {
         return *config;
     }
 
 private:
     std::shared_ptr<Everest::RuntimeSettings> rs;
```
a-w50: no shared pointer!

include/framework/everest.hpp
```diff
@@ -48,9 +48,8 @@ struct ErrorFactory;
 class Everest {
 public:
     Everest(std::string module_id, const Config& config, bool validate_data_with_schema,
-            const std::string& mqtt_server_socket_path, const std::string& mqtt_server_address, int mqtt_server_port,
-            const std::string& mqtt_everest_prefix, const std::string& mqtt_external_prefix,
-            const std::string& telemetry_prefix, bool telemetry_enabled);
+            const std::shared_ptr<MQTTSettings> mqtt_settings, const std::string& telemetry_prefix,
```
a-w50: `const std::shared_ptr` doesn't fit, `std::shared_ptr<const MQTTSettings>` probably would, but still, don't use shared pointer here, either pass by value or by const ref (the lifetime of the settings should be longer than the lifetime of the `Everest` instance I guess).

include/framework/runtime.hpp
```diff
@@ -12,6 +12,7 @@
 #include <framework/ModuleAdapter.hpp>
 #include <sys/prctl.h>
```
a-w50, line 11: include order is a bit messed up here, `sys/prctl.h` should go up and `framework` and `utils` belong togeter.  We should probably create an issue for fixing the include directory namespacing in everest framework

include/framework/runtime.hpp
```diff
@@ -91,47 +92,65 @@ const auto TERMINAL_STYLE_ERROR = fmt::emphasis::bold | fg(fmt::terminal_color::
...
-    std::string mqtt_broker_socket_path;
-    std::string mqtt_broker_host;
-    int mqtt_broker_port;
-    std::string mqtt_everest_prefix;
-    std::string mqtt_external_prefix;
-    std::string telemetry_prefix;
-    bool telemetry_enabled;
+/// \brief Settings needed by the manager to load and validate a config
+
+struct ManagerSettings {
```
a-w50, line 127: It is a good idea to split up the manager settings and the runtime settings, but the way it is done here, it leads to some code duplication.  The `ManagerSettings` does also create the `RuntimeSettings`, so why not just make `RuntimeSettings` a part of `ManagerSettings`?

include/framework/runtime.hpp
```diff
@@ -146,19 +165,24 @@ struct ModuleCallbacks {
                     const std::function<void()>& ready);
 };
 
+///\brief Version information
 struct VersionInformation {
-    std::string project_name;
-    std::string project_version;
-    std::string git_version;
+    std::string project_name;    ///< CMake project name
+    std::string project_version; ///< Human readable version number
```
a-w50, line 174: Not sure if `Human` should be written in lower case as it is not a full sentence

include/framework/runtime.hpp
```diff
@@ -146,19 +165,24 @@ struct ModuleCallbacks {
...
+    std::string git_version;     ///< Git version containing tag and branch information
 };
 
 class ModuleLoader {
 private:
     std::shared_ptr<RuntimeSettings> runtime_settings;
+    std::shared_ptr<MQTTSettings> mqtt_settings;
```
a-w50: no shared pointer

include/utils/config.hpp
```diff
@@ -256,21 +336,39 @@ class Config {
     /// \brief Extracts the keys of the provided json \p object
     ///
     /// \returns a set of object keys
-    static std::set<std::string> keys(json object);
+    static std::set<std::string> keys(const json& object);
```
a-w50: should be a free function, not a static member function
same goes for all the other static member functions

include/utils/config.hpp
```diff
@@ -43,50 +46,27 @@ struct schemas {
 const static std::regex type_uri_regex{R"(^((?:\/[a-zA-Z0-9\-\_]+)+#\/[a-zA-Z0-9\-\_]+)$)"};
 
 ///
-/// \brief Contains config and manifest parsing
+/// \brief Base class for configs
```
a-w50, line 54: *inheritance is the base class of evil* (https://www.youtube.com/watch?v=2bLkxj6EVoM)
This doesn't look right to me.  I think this should be solved using composition

include/utils/config.hpp
```diff
@@ -203,34 +178,139 @@ class Config {
...
+
+    ///
+    /// \brief creates a new Config object form the given \p mqtt_settings and \p config
+    explicit Config(std::shared_ptr<MQTTSettings> mqtt_settings, json config);
+
+    ///
+    /// \brief checks if the given \p module_id provides the requirement given in \p requirement_id
+    ///
+    /// \returns a json object that contains the requirement
+    json resolve_requirement(const std::string& module_id, const std::string& requirement_id) const;
```
a-w50: Why for example should a module need to resolve a requirement.  This is clearly the managers responsibility

include/utils/config_cache.hpp
```diff
@@ -18,5 +18,17 @@ struct ConfigCache {
...
+    static void to_json(nlohmann::json& j, const Everest::ConfigCache& c) {
+        j = {{"provides_impl", c.provides_impl}, {"cmds", c.cmds}};
+    }
+
+    static void from_json(const nlohmann::json& j, Everest::ConfigCache& c) {
+        c.provides_impl = j.at("provides_impl");
+        c.cmds = j.at("cmds");
+    }
+};
+NLOHMANN_JSON_NAMESPACE_END
```
a-w50: btw, these functions should be forward declared and implemented in a cpp file.

include/utils/module_config.hpp
```diff
@@ -0,0 +1,49 @@
...
+/// \brief Check if the provided \p path is a file
+/// \returns The canonical version of the provided path
+/// \throws BootException if the path doesn't exist or isn't a regular file
+fs::path assert_file(const std::string& path, const std::string& file_alias = "The");
+
+/// \returns true if the file at the provided \p path has an extensions \p ext
+bool has_extension(const std::string& path, const std::string& ext);
+
+/// \returns a path that has been prefixed by \p prefix from the provided json \p value
+std::string get_prefixed_path_from_json(const nlohmann::json& value, const fs::path& prefix);
```
a-w50: These functions are not really related to `module_config`, they should be moved to a *more* common header file

include/utils/module_config.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+///
+/// \brief Contains helpers for the module config loading
+///
+class ModuleConfig {
+public:
+    ModuleConfig() = default;
+
+    /// \brief get config from manager via mqtt
+    static json get_config(std::shared_ptr<MQTTSettings> mqtt_settings, const std::string& module_id);
```
a-w50: This class has only one static function.  So it doesn't make sense to have this class at all.  Just make a free function out of `get_config`.

include/utils/module_config.hpp
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Pionix GmbH and Contributors to EVerest
+#ifndef UTILS_MODULE_CONFIG_HPP
+#define UTILS_MODULE_CONFIG_HPP
+
+#include <future>
+
+#include <nlohmann/json.hpp>
+
+#include <utils/mqtt_abstraction.hpp>
```
a-w50, line 8: Why should `module_config` need to know about the `mqtt_abstraction`?

include/utils/module_config.hpp
a-w50: consistency, here you use `nlohmann::json`, above just `json`.  In general it is a good idea to use full namespaces in header files and use `using` for shortening (in contrast to giving meaning) in the cpp files

include/utils/mqtt_abstraction.hpp
```diff
@@ -15,14 +15,34 @@ using json = nlohmann::json;
 // forward declaration
 class MQTTAbstractionImpl;
 
+/// \brief minimal MQTT connection settings needed for an initial connection of a module to the manager
+struct MQTTSettings {
```
a-w50: This class clearly belongs into its own header file, this will also solve the problem, mentioned above.
Mqtt settings don't need to be aware of the interface how you access the mqtt bus.

include/utils/mqtt_abstraction_impl.hpp
```diff
@@ -59,15 +59,15 @@ class MQTTAbstractionImpl {
...
+    void publish(const std::string& topic, const json& json, QOS qos, bool retain = false);
 
     ///
     /// \brief publishes the given \p data on the given \p topic with QOS level 0
     void publish(const std::string& topic, const std::string& data);
 
     ///
     /// \brief publishes the given \p data on the given \p topic with the given \p qos
-    void publish(const std::string& topic, const std::string& data, QOS qos);
+    void publish(const std::string& topic, const std::string& data, QOS qos, bool retain = false);
```
a-w50, line 69: This `retain` behavior better might be in a different commit

include/utils/types.hpp
```diff
@@ -39,6 +39,7 @@ enum class HandlerType {
     SubscribeVar,
     SubscribeError,
     ClearErrorRequest,
+    Internal,
```
a-w50: `Internal` is really broad, could be everything.

include/utils/types.hpp
```diff
@@ -82,6 +83,8 @@ struct ModuleInfo {
 
 struct TelemetryConfig {
     int id;
+    explicit TelemetryConfig(int id) : id(id) {
+    }
```
a-w50, line 134: standard list initialization should be enough.  This struct is also a bit redundant.  Something like `using TelemetryConfigID = int` would probably have been enough

include/utils/types.hpp
```diff
@@ -138,6 +141,44 @@ template <> struct adl_serializer<Mapping> {
...
+            }
+            if (j.contains("implementations")) {
+                for (auto& impl : j.at("implementations").items()) {
+                    m.implementations[impl.key()] = impl.value();
+                }
+            }
+        }
+        return m;
+    }
+};
```
a-w50: implementation should go into a cpp file

lib/module_config.cpp
```diff
@@ -0,0 +1,159 @@
...
+    std::promise<json> res_promise;
+    std::future<json> res_future = res_promise.get_future();
+
+    Handler res_handler = [module_id, &res_promise](json data) {
+        EVLOG_verbose << fmt::format("Incoming config for {}", module_id);
+
+        res_promise.set_value(std::move(data));
+    };
+
+    std::shared_ptr<TypedHandler> res_token =
```
a-w50: no shared pointer

lib/module_config.cpp
```diff
@@ -0,0 +1,159 @@
...
+std::string get_prefixed_path_from_json(const nlohmann::json& value, const fs::path& prefix) {
+    auto settings_configs_dir = value.get<std::string>();
+    if (fs::path(settings_configs_dir).is_relative()) {
+        settings_configs_dir = (prefix / settings_configs_dir).string();
+    }
+    return settings_configs_dir;
+}
+
+json ModuleConfig::get_config(std::shared_ptr<MQTTSettings> mqtt_settings, const std::string& module_id) {
+    auto mqtt = MQTTAbstraction(mqtt_settings);
```
a-w50: It should not be the `ModuleConfig`s responsibility to spawn an mqtt client thread in order to retreive its configuration.  The mqtt client will be necessary later on anyway for communication, why not use that instance to retreive the configuration?

lib/module_config.cpp
```diff
@@ -0,0 +1,159 @@
...
+    auto manifests_topic = fmt::format("{}manifests", mqtt_settings->mqtt_everest_prefix);
+    auto manifests = mqtt.get(manifests_topic, QOS::QOS2);
+    result["manifests"] = manifests;
+
+    auto error_types_map_topic = fmt::format("{}error_types_map", mqtt_settings->mqtt_everest_prefix);
+    auto error_types_map = mqtt.get(error_types_map_topic, QOS::QOS2);
+    result["error_map"] = error_types_map;
+
+    auto module_config_cache_topic = fmt::format("{}module_config_cache", mqtt_settings->mqtt_everest_prefix);
+    auto module_config_cache = mqtt.get(module_config_cache_topic, QOS::QOS2);
```
a-w50: const

lib/module_config.cpp
```diff
@@ -0,0 +1,159 @@
...
+    } while (res_future_status == std::future_status::deferred);
+
+    json result;
+    if (res_future_status == std::future_status::timeout) {
+        EVLOG_AND_THROW(
+            EverestTimeoutError(fmt::format("Timeout while waiting for result of get_config of {}", module_id)));
+    } else if (res_future_status == std::future_status::ready) {
+        EVLOG_verbose << "res future ready";
+        result = res_future.get();
+    }
```
a-w50: If having an `if` containing a `throw` , you don't need to do an `else if`.  Also a log like `res future ready` will probably help no one.

lib/config.cpp
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright 2020 - 2023 Pionix GmbH and Contributors to EVerest
+// Copyright Pionix GmbH and Contributors to EVerest
```
a-w50: This file is just to much for me to reason about.  I think the chosen abstraction should be improved and a clear statement about the responsibilities of `ManagerConfig`, `ConfigBase`, `Config` and `ModuleConfig`   should be done for  in the documentation

lib/config.cpp
```diff
@@ -790,33 +852,63 @@ ModuleConfigs Config::get_module_configs(const std::string& module_id) const {
     return module_configs;
 }
 
-const json& Config::get_manifests() {
+const json& ConfigBase::get_manifests() {
```
a-w50: function should be const

include/utils/error/error_type_map.hpp
```diff
@@ -50,6 +56,12 @@ class ErrorTypeMap {
     ///
     bool has(const ErrorType& error_type) const;
 
+    ///
+    /// \brief Returns the contained ErrorType map
+    /// \return The error types map
+    ///
+    std::map<ErrorType, std::string> get_error_types_map();
```
a-w50: naming!  `ErrorTypeMap` is a map of probably multiple error types, therefore it is okay to name the private property `error_types`.  But then the function can't be called `get_error_types_map`, either `get_error_types` or `get_error_type_map` (but for the latter one you could also just write `get` as the class is already called `ErrorTypeMap`.  Also note that with this function you expose the full private data of this class, rendering it useless with respect to *privacy* or *implementation details*.  This whole thing probably just wants to be a `using ErrorTypeMap = std::map<ErrorType, std::string>` with free functions like
```
ErrorTypeMap load_error_types(..);
/* load_error_types_map will not be needed */
std::string get_description(const ErrorTypeMap&, const ErrorType&);
bool has(const ErrorTypeMap&, const ErrorType&);
/* get_error_types_map is not needed */
```

include/utils/module_config.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <utils/types.hpp>
+
+namespace Everest {
+using json = nlohmann::json;
+namespace fs = std::filesystem;
+
+struct BootException : public std::runtime_error {
```
a-w50: why is `BootException` inside `module_config`?  Either namespace it or prefix it which `Module` because catching `BootException` somewhere in the code will not make easily clear that this belongs to the startup of a module.

include/utils/mqtt_abstraction.hpp
```diff
@@ -15,14 +15,34 @@ using json = nlohmann::json;
...
+    std::string mqtt_broker_socket_path; ///< A path to a socket the MQTT broker uses in socket mode. If this is set
+                                         ///< mqtt_broker_host and mqtt_broker_port are ignored
+    std::string mqtt_broker_host;        ///< The hostname of the MQTT broker
+    int mqtt_broker_port = 0;            ///< The port the MQTT broker listens on
+    std::string mqtt_everest_prefix;     ///< MQTT topic prefix for the "everest" topic
+    std::string mqtt_external_prefix;    ///< MQTT topic prefix for external topics
+    bool socket = false; ///< Indicates if a Unix Domain Socket is used for connection to the MQTT broker
```
a-w50: this could probably be a function, checking if `mqtt_broker_socket_path` is not empty.  Furthermore, I wouldn't prefix all the properties with mqtt, because the struct is already called `MqttSettings`

include/utils/mqtt_abstraction.hpp
```diff
@@ -15,14 +15,34 @@ using json = nlohmann::json;
...
 ///
 /// \brief Contains a C++ abstraction for using MQTT in EVerest modules
 ///
 class MQTTAbstraction {
 public:
-    MQTTAbstraction(const std::string& mqtt_server_socket_path, const std::string& mqtt_server_address,
-                    const std::string& mqtt_server_port, const std::string& mqtt_everest_prefix,
-                    const std::string& mqtt_external_prefix);
+    /// \brief Create a MQTTAbstraction with the provideded \p mqtt_settings
+    MQTTAbstraction(const std::shared_ptr<MQTTSettings> mqtt_settings);
```
a-w50: no shared pointer

lib/everest.cpp
```diff
@@ -33,17 +33,15 @@ const auto remote_cmd_res_timeout_seconds = 300;
 const std::array<std::string, 3> TELEMETRY_RESERVED_KEYS = {{"connector_id"}};
 
 Everest::Everest(std::string module_id_, const Config& config_, bool validate_data_with_schema,
-                 const std::string& mqtt_server_socket_path, const std::string& mqtt_server_address,
-                 int mqtt_server_port, const std::string& mqtt_everest_prefix, const std::string& mqtt_external_prefix,
-                 const std::string& telemetry_prefix, bool telemetry_enabled) :
-    mqtt_abstraction(mqtt_server_socket_path, mqtt_server_address, std::to_string(mqtt_server_port),
-                     mqtt_everest_prefix, mqtt_external_prefix),
+                 const std::shared_ptr<MQTTSettings> mqtt_settings, const std::string& telemetry_prefix,
```
a-w50: no shared pointers!

lib/everest.cpp
```diff
@@ -520,11 +519,12 @@ void Everest::subscribe_error(const Requirement& req, const error::ErrorType& er
...
     std::string module_name = this->config.get_module_name(requirement_module_id);
     std::string requirement_impl_id = connection.at("implementation_id");
-    json requirement_impl_if = this->config.get_interfaces().at(module_name).at(requirement_impl_id);
+    json requirement_impl_if = this->config.get_interface_definitions().at(
+        this->config.get_interfaces().at(module_name).at(requirement_impl_id));
 
     // check if requirement is allowed to publish this error_type
     // split error_type at '/'
-    int pos = error_type.find('/');
+    size_t pos = error_type.find('/');
```
a-w50: just use `const auto`

lib/mqtt_abstraction.cpp
```diff
@@ -1,21 +1,38 @@
...
+        mqtt_abstraction = std::make_unique<MQTTAbstractionImpl>(mqtt_settings->mqtt_broker_socket_path,
+                                                                 mqtt_settings->mqtt_everest_prefix,
+                                                                 mqtt_settings->mqtt_external_prefix);
     } else {
-        mqtt_abstraction =
-            std::make_unique<MQTTAbstractionImpl>(mqtt_server_socket_path, mqtt_everest_prefix, mqtt_external_prefix);
+        mqtt_abstraction = std::make_unique<MQTTAbstractionImpl>(
+            mqtt_settings->mqtt_broker_host, std::to_string(mqtt_settings->mqtt_broker_port),
+            mqtt_settings->mqtt_everest_prefix, mqtt_settings->mqtt_external_prefix);
     }
```
a-w50: always try to prefer direct initialization.  Here you can just write a simple factory method like
```
auto create_mqtt_client(const MQTTSettings& sessings) {
  if (settings.socket) {
    return std::make_unique<...>(..);
  } else {
    return std::make_unique<...>(..);
  }
}
```
And then the constructor will look like
```
MQTTAbstraction::MQTTAbstraction(const MQTTSettings& settings) : mqtt_abstraction(create_mqtt_client(settings)) {}
```
Note that the helper function `create_mqtt_client` can also be easily reused in tests and also properly tested by itself!

lib/mqtt_abstraction_impl.cpp
```diff
@@ -198,6 +202,39 @@ void MQTTAbstractionImpl::unsubscribe(const std::string& topic) {
...
 }
 
+json MQTTAbstractionImpl::get(const std::string& topic, QOS qos) {
+    BOOST_LOG_FUNCTION();
+    std::promise<json> res_promise;
+    std::future<json> res_future = res_promise.get_future();
+
+    Handler res_handler = [this, &res_promise](json data) { res_promise.set_value(std::move(data)); };
+
+    std::shared_ptr<TypedHandler> res_token =
```
a-w50: no shared pointer!

lib/mqtt_abstraction_impl.cpp
```diff
@@ -198,6 +202,39 @@ void MQTTAbstractionImpl::unsubscribe(const std::string& topic) {
...
+        std::make_shared<TypedHandler>(HandlerType::Internal, std::make_shared<Handler>(res_handler));
+    this->register_handler(topic, res_token, QOS::QOS2);
+
+    json config_publish_data = json::object({{"type", "full"}});
+
+    // wait for result future
+    std::chrono::time_point<std::chrono::steady_clock> res_wait =
+        std::chrono::steady_clock::now() + std::chrono::seconds(10);
```
a-w50: hardcoded value -> explicit constant

lib/runtime.cpp
```diff
@@ -348,6 +329,26 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string&
...
+    } else {
+        this->mqtt_settings = std::make_shared<MQTTSettings>(mqtt_broker_host, mqtt_broker_port, mqtt_everest_prefix,
+                                                             mqtt_external_prefix);
+    }
+
+    config = config_;
+
+    run_as_user = settings.value("run_as_user", "");
+
+    auto version_information_path = data_dir / "version_information.txt";
```
a-w50: hardcoded string literal -> explicit constant

lib/runtime.cpp
```diff
@@ -269,6 +244,12 @@ RuntimeSettings::RuntimeSettings(const std::string& prefix_, const std::string&
         controller_rpc_timeout_ms = defaults::CONTROLLER_RPC_TIMEOUT_MS;
     }
 
+    std::string mqtt_broker_socket_path;
+    std::string mqtt_broker_host;
+    int mqtt_broker_port = 0;
+    std::string mqtt_everest_prefix;
+    std::string mqtt_external_prefix;
+
```
a-w50, line 257: what is this for?

lib/runtime.cpp
```diff
@@ -392,29 +390,53 @@ ModuleCallbacks::ModuleCallbacks(const std::function<void(ModuleAdapter module_a
...
-    runtime_settings(nullptr), callbacks(callbacks), version_information(version_information) {
+    runtime_settings(nullptr), mqtt_settings(nullptr), callbacks(callbacks), version_information(version_information) {
     if (!this->parse_command_line(argc, argv)) {
+        this->should_exit = true;
         return;
     }
 }
 
 int ModuleLoader::initialize() {
+    if (this->should_exit) {
```
a-w50, line 409: If one needs to check for `should_exit` in an `initialize` function, probably the class' construction is not managed in a proper way.

lib/runtime.cpp
```diff
@@ -563,9 +598,97 @@ bool ModuleLoader::parse_command_line(int argc, char* argv[]) {
...
+        }
+    }
+
+    if (vm.count("mqtt_broker_port") != 0) {
+        mqtt_broker_port = vm["mqtt_broker_port"].as<int>();
+    }
+
+    // overwrite mqtt broker port with environment variable
+    // NOLINTNEXTLINE(concurrency-mt-unsafe): not problematic that this function is not threadsafe here
+    const char* mqtt_server_port = std::getenv("MQTT_SERVER_PORT");
```
a-w50: isn't this code duplicate to the one in the everestpy.cpp for example?

src/manager.cpp
```diff
@@ -237,26 +274,110 @@ static std::map<pid_t, std::string> spawn_modules(const std::vector<ModuleStartI
...
+    auto manifests = config.get_manifests();
+    mqtt_abstraction.publish(fmt::format("{}manifests", ms->mqtt_settings->mqtt_everest_prefix), manifests, QOS::QOS2,
+                             true);
+
+    auto error_types_map = config.get_error_types_map();
+    mqtt_abstraction.publish(fmt::format("{}error_types_map", ms->mqtt_settings->mqtt_everest_prefix), error_types_map,
+                             QOS::QOS2, true);
+
+    auto module_config_cache = config.get_module_config_cache();
+    mqtt_abstraction.publish(fmt::format("{}module_config_cache", ms->mqtt_settings->mqtt_everest_prefix),
```
a-w50: What are these topics for?  Who is going to use them?

tests/test_config.cpp
```diff
@@ -8,99 +8,206 @@
...
+        THEN("It should not throw at all") {
+            CHECK_NOTHROW([&]() {
+                auto mc = Everest::ManagerConfig(rs);
+                auto main = mc.get_main_config();
+                CHECK(main.at("valid_module").at("config_module").at("valid_config_entry") == "hi");
+            }());
+        }
+    }
+    GIVEN("A valid config with a valid module and enabled schema validation") {
+        std::shared_ptr<Everest::ManagerSettings> rs =
```
a-w50: use auto

include/utils/module_config.hpp
hikinggrass: I've reduced the usage of this in multiple headers

include/utils/config.hpp
hikinggrass: lets refactor this in a followup PR

include/utils/error/error_type_map.hpp
hikinggrass: Renamed the funtion to get_error_types. Maybe this can be refactored further in the future

CMakeLists.txt
```diff
@@ -30,9 +30,11 @@ ev_setup_python_executable(
...
 if((${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME} OR ${PROJECT_NAME}_BUILD_TESTING) AND BUILD_TESTING)
     set(EVEREST_FRAMEWORK_BUILD_TESTING ON)
-    evc_include(CodeCoverage)
-
-    append_coverage_compiler_flags()
+    # this policy allows us to link gcov to targets defined in other directories
+    if(POLICY CMP0079)
+        cmake_policy(SET CMP0079 NEW)
+        set(CMAKE_POLICY_DEFAULT_CMP0079 NEW)
+    endif()
```
a-w50: We shouldn't need this, as the required cmake version is already > 3.13.
Furthermore, `set(CMAKE_POLICY_DEFAULT_CMP0079 NEW)` already enforces `cmake_policy(SET CMP0079 NEW)`, so only the latter would be necessary.  I also don't get the point of `if(POLICY CMP0079)`

lib/everest.cpp
```diff
@@ -42,7 +42,7 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
                  std::shared_ptr<MQTTAbstraction> mqtt_abstraction, const std::string& telemetry_prefix,
                  bool telemetry_enabled) :
     mqtt_abstraction(mqtt_abstraction),
-    config(std::move(config_)),
```
a-w50: moving a `const` is indeed pointless.  But it might be a better choice to make the constructor argument `Config config_` instead of const ref and dropping the move.

lib/everest.cpp
```diff
@@ -188,8 +188,8 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
     }
 
     // register handler for global ready signal
-    const Handler handle_ready_wrapper = [this](json data) { this->handle_ready(data); };
-    const std::shared_ptr<TypedHandler> everest_ready =
```
a-w50: why have all the `const`s disappeared?  `auto` can also be used for the `shared_ptr` variables

a-w50 (approved):

CMakeLists.txt
hikinggrass: I'm also very confused why this is needed, since it should be the default. However on CMake 3.31 I get a
```
Attempt to add link library "gcov" to target "manager" which is not built
  in this directory.

  This is allowed only when policy CMP0079 is set to NEW.
Call Stack (most recent call first):
  tests/CMakeLists.txt:69 (append_coverage_compiler_flags_to_target)
```
error as if that policy isn't set.
the `if(POLICY CMP0079)` just ensures that the policy is available before setting it, maybe not necessary on CMake >= 3.13

lib/everest.cpp
hikinggrass: Looks like that's from the recent merge of main where this turned from `Handler` to `auto`. I'll change that

PR #206, opened 2024-10-06
Expose mappings of requirements, provides and modules to modules

### LINKED ISSUE #199: Make 3-tier mapping available in Module Info
Currently the 3-tier mapping can be used in the everest config file and is part of the errors that are raised.
The goal of this issue is to make the mapping available as part of the ModuleInfo. This would lead to the fact that some modules dont need additional config parameters like `evse_id` or `connector_id` (e.g. the EvseManager)

hikinggrass:
EVerest modules and even individual interface implementations can have mappings
assigned to them. These mappings are inspired by the OCPP 3-tier model.

Modules can access the mapping information in the following ways depending on which specific information is required, all of these return an optional `Mapping` struct
If the mapping of a requirement is of interest it can be accessed via a get_mapping() function of a requirments:
`r_name_of_the_requirement->get_mapping()`

If the mapping of an interface implementation is of interest it can also be accessed via a get_mapping() function: `p_name_of_an_implementation->get_mapping()`

If the mapping of the current module is of interest it can be accessed via the module info:
`this->info.mapping`

Mapping information is also available in error reporting via "error.origin.mapping":
```
const auto error_handler = [this](const Everest::error::Error& error) {
    const auto evse_id = error.origin.mapping.has_value() ? error.origin.mapping.value().evse : 0;
};
```

### DIFF
--- a/include/framework/ModuleAdapter.hpp
+++ b/include/framework/ModuleAdapter.hpp
@@ -93,7 +93,6 @@ struct ModuleAdapter {
     using TelemetryPublishFunc =
         std::function<void(const std::string&, const std::string&, const std::string&, const TelemetryMap&)>;
     using GetMappingFunc = std::function<std::optional<ModuleTierMappings>()>;
-    using GetImplMappingFunc = std::function<std::optional<Mapping>(const std::string&)>;
 
     CallFunc call;
     PublishFunc publish;
@@ -110,7 +109,6 @@ struct ModuleAdapter {
     std::vector<cmd> registered_commands;
     TelemetryPublishFunc telemetry_publish;
     GetMappingFunc get_mapping;
-    GetImplMappingFunc get_impl_mapping;
 
     void check_complete() {
         // FIXME (aw): I should throw if some of my handlers are not set
--- a/include/framework/ModuleAdapter.hpp
+++ b/include/framework/ModuleAdapter.hpp
@@ -92,6 +92,8 @@ struct ModuleAdapter {
     using ExtMqttSubscribeFunc = std::function<UnsubscribeToken(const std::string&, StringHandler)>;
     using TelemetryPublishFunc =
         std::function<void(const std::string&, const std::string&, const std::string&, const TelemetryMap&)>;
+    using GetMappingFunc = std::function<std::optional<ModuleTierMappings>()>;
+    using GetImplMappingFunc = std::function<std::optional<Mapping>(const std::string&)>;
 
     CallFunc call;
     PublishFunc publish;
@@ -107,6 +109,8 @@ struct ModuleAdapter {
     ExtMqttSubscribeFunc ext_mqtt_subscribe;
     std::vector<cmd> registered_commands;
     TelemetryPublishFunc telemetry_publish;
+    GetMappingFunc get_mapping;
+    GetImplMappingFunc get_impl_mapping;
 
     void check_complete() {
         // FIXME (aw): I should throw if some of my handlers are not set
--- a/include/framework/everest.hpp
+++ b/include/framework/everest.hpp
@@ -148,11 +148,6 @@ public:
     std::optional<ModuleTierMappings> get_3_tier_model_mapping();
 
     ///
-    /// \returns the 3 tier model mapping for the given \p impl_id
-    ///
-    std::optional<Mapping> get_3_tier_model_mapping(const std::string& impl_id);
-
-    ///
     /// \brief Chccks if all commands of a module that are listed in its manifest are available
     ///
     void check_code();
@@ -249,6 +244,13 @@ private:
     ///
     void subscribe_global_all_errors(const error::ErrorCallback& callback, const error::ErrorCallback& clear_callback);
 };
+
+///
+/// \returns the 3 tier model mapping from a \p module_tier_mapping for the given \p impl_id
+///
+std::optional<Mapping> get_impl_mapping(std::optional<ModuleTierMappings> module_tier_mappings,
+                                        const std::string& impl_id);
+
 } // namespace Everest
 
 #endif // FRAMEWORK_EVEREST_HPP
--- a/include/framework/everest.hpp
+++ b/include/framework/everest.hpp
@@ -143,6 +143,16 @@ public:
     bool is_telemetry_enabled();
 
     ///
+    /// \returns the 3 tier model mappings for this module
+    ///
+    std::optional<ModuleTierMappings> get_3_tier_model_mapping();
+
+    ///
+    /// \returns the 3 tier model mapping for the given \p impl_id
+    ///
+    std::optional<Mapping> get_3_tier_model_mapping(const std::string& impl_id);
+
+    ///
     /// \brief Chccks if all commands of a module that are listed in its manifest are available
     ///
     void check_code();
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -147,8 +147,14 @@ public:
     json resolve_requirement(const std::string& module_id, const std::string& requirement_id) const;
 
     ///
-    /// \returns a map of Fulfillments for \p module_id
+    /// \returns a list of Requirements for \p module_id
     ///
+    std::list<Requirement> get_requirements(const std::string& module_id) const;
+
+    ///
+    /// \brief A Fulfillment is a combination of a Requirement and the module and implementation ids where this is
+    /// implemented
+    /// \returns a map of Fulfillments for \p module_id
     std::map<std::string, std::vector<Fulfillment>> get_fulfillments(const std::string& module_id);
 
     ///
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -105,12 +105,17 @@ private:
 
     ///
     /// \brief Parses the 3 tier model mappings in the config
+    /// A "mapping" can be specified in the following way:
     /// You can set a EVSE id called "evse" and Connector id called "connector" for the whole module.
-    /// Additionally a "mapping" can be specified in the following way:
+    /// Alternatively you can set individual mappings for implementations.
     /// mapping:
-    ///   implementation_id:
+    ///   module:
     ///     evse: 1
     ///     connector: 1
+    ///   implementations:
+    ///     implementation_id:
+    ///       evse: 1
+    ///       connector: 1
     /// If no mappings are found it will be assumed that the module is mapped to the charging station.
     /// If only a module mapping is defined alle implementations are mapped to this module mapping.
     /// Implementations can have overwritten mappings.
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -147,9 +147,9 @@ public:
     json resolve_requirement(const std::string& module_id, const std::string& requirement_id) const;
 
     ///
-    /// \returns a map of RequirementConnections for \p module_id
+    /// \returns a map of Fulfillments for \p module_id
     ///
-    std::map<std::string, std::vector<RequirementConnection>> get_requirement_connections(const std::string& module_id);
+    std::map<std::string, std::vector<Fulfillment>> get_fulfillments(const std::string& module_id);
 
     ///
     /// \brief checks if the config contains the given \p module_id
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -204,7 +204,7 @@ public:
 
     //
     /// \returns the 3 tier model mappings for the given \p module_id
-    std::optional<ModuleTierMappings> get_3_tier_model_mappings(const std::string& module_id);
+    std::optional<ModuleTierMappings> get_module_3_tier_model_mappings(const std::string& module_id);
 
     //
     /// \returns the 3 tier model mapping for the given \p module_id and \p impl_id
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -147,6 +147,13 @@ public:
     json resolve_requirement(const std::string& module_id, const std::string& requirement_id) const;
 
     ///
+    /// \brief resolves all Requirements of the given \p module_id to their Fulfillments
+    ///
+    /// \returns a map indexed by Requirements
+    std::map<Requirement, Fulfillment> resolve_requirements(const std::string& module_id) const;
+
+
+    ///
     /// \returns a list of Requirements for \p module_id
     ///
     std::list<Requirement> get_requirements(const std::string& module_id) const;
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -152,7 +152,6 @@ public:
     /// \returns a map indexed by Requirements
     std::map<Requirement, Fulfillment> resolve_requirements(const std::string& module_id) const;
 
-
     ///
     /// \returns a list of Requirements for \p module_id
     ///
@@ -163,6 +162,12 @@ public:
     /// implemented
     /// \returns a map of Fulfillments for \p module_id
     std::map<std::string, std::vector<Fulfillment>> get_fulfillments(const std::string& module_id) const;
+
+    ///
+    /// \brief A RequirementInitialization contains everything needed to initialize a requirement in user code. This
+    /// includes the Requirement, its Fulfillment and an optional Mapping
+    /// \returns a RequirementInitialization
+    RequirementInitialization get_requirement_initialization(const std::string& module_id) const;
 
     ///
     /// \brief checks if the config contains the given \p module_id
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -147,9 +147,9 @@ public:
     json resolve_requirement(const std::string& module_id, const std::string& requirement_id) const;
 
     ///
-    /// \returns a list of Requirements for \p module_id
+    /// \returns a map of RequirementConnections for \p module_id
     ///
-    std::list<Requirement> get_requirements(const std::string& module_id) const;
+    std::map<std::string, std::vector<RequirementConnection>> get_requirement_connections(const std::string& module_id);
 
     ///
     /// \brief checks if the config contains the given \p module_id
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -155,7 +155,7 @@ public:
     /// \brief A Fulfillment is a combination of a Requirement and the module and implementation ids where this is
     /// implemented
     /// \returns a map of Fulfillments for \p module_id
-    std::map<std::string, std::vector<Fulfillment>> get_fulfillments(const std::string& module_id);
+    std::map<std::string, std::vector<Fulfillment>> get_fulfillments(const std::string& module_id) const;
 
     ///
     /// \brief checks if the config contains the given \p module_id
@@ -204,11 +204,11 @@ public:
 
     //
     /// \returns the 3 tier model mappings for the given \p module_id
-    std::optional<ModuleTierMappings> get_module_3_tier_model_mappings(const std::string& module_id);
+    std::optional<ModuleTierMappings> get_module_3_tier_model_mappings(const std::string& module_id) const;
 
     //
     /// \returns the 3 tier model mapping for the given \p module_id and \p impl_id
-    std::optional<Mapping> get_3_tier_model_mapping(const std::string& module_id, const std::string& impl_id);
+    std::optional<Mapping> get_3_tier_model_mapping(const std::string& module_id, const std::string& impl_id) const;
 
     ///
     /// \brief turns then given \p module_id into a printable identifier
--- a/include/utils/types.hpp
+++ b/include/utils/types.hpp
@@ -138,13 +138,22 @@ struct Requirement {
 };
 
 /// \brief A Fulfillment relates a Requirement to its connected implementation, identified via its module and
-/// implementation id. It additionally provides access to the Mapping associated with the Requirement.
+/// implementation id.
 struct Fulfillment {
     std::string module_id;
     std::string implementation_id;
     Requirement requirement;
     std::optional<Mapping> mapping;
 };
+
+/// \brief Contains everything that's needed to initialize a requirement in user code
+struct RequirementInitializer {
+    Requirement requirement;
+    Fulfillment fulfillment;
+    std::optional<Mapping> mapping;
+};
+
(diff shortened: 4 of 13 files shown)


Pietfried (approved):
lgtm. Linting is currently failing

Tested with https://github.com/EVerest/everest-core/pull/872

include/utils/types.hpp
```diff
@@ -96,6 +76,30 @@ struct Mapping {
...
+    return os;
+}
+
+/// \brief Writes the string representation of the given Mapping \p mapping to the given output stream \p os
+/// \returns an output stream with the Mapping written to
+inline std::ostream& operator<<(std::ostream& os, const std::optional<Mapping>& mapping) {
+    if (mapping.has_value()) {
+        os << mapping.value();
+    } else {
+        os << "Mapping(charging station)";
```
a-w50, line 97: Does this default representation make any sense?

include/utils/types.hpp
```diff
@@ -96,6 +76,30 @@ struct Mapping {
...
+        os << ", connector: " << mapping.connector.value();
+    }
+    os << ")";
+
+    return os;
+}
+
+/// \brief Writes the string representation of the given Mapping \p mapping to the given output stream \p os
+/// \returns an output stream with the Mapping written to
+inline std::ostream& operator<<(std::ostream& os, const std::optional<Mapping>& mapping) {
```
a-w50, line 93: The stream output operator should probably be defined for `Mapping` and not `std::optional<Mapping>`

include/utils/types.hpp
```diff
@@ -104,11 +108,39 @@ struct ModuleTierMappings {
...
+    Requirement() = default;
+    Requirement(const std::string& requirement_id_, size_t index_, std::optional<Mapping> mapping = std::nullopt);
     bool operator<(const Requirement& rhs) const;
     std::string id;
-    size_t index;
+    size_t index = 0;
+    std::optional<Mapping> mapping;
+};
+
+struct RequirementConnection {
```
a-w50: What is a `RequirementConnection`?  Maybe there's a better word for this.  Maybe it is a fulfillment?

a-w50 (changes requested):
I would like to request some documentation in the commit message about what has been implemented here.

hikinggrass:
> I would like to request some documentation in the commit message about what has been implemented here.

Added to the PR description which will be in the squashed commit message, it's also already available in the documentation: https://everest.github.io/nightly/general/05_existing_modules.html#tier-module-mappings (the code snippets didn't render, should be fixed soon)

include/utils/types.hpp
hikinggrass: The feature (at the moment at least) is inspired by the 3-tier mappings from OCPP. So if you have no explicit mapping it's assumed that the module maps to the whole charging station, getting more specific by a mapping to an evse and/or connector

include/utils/types.hpp
hikinggrass: It already is: https://github.com/EVerest/everest-framework/blob/557932c30108ce09b3670f01cb49c5d5092b681f/include/utils/types.hpp#L81

include/utils/types.hpp
hikinggrass: Renamed to Fulfillment as already used by everestpy

include/framework/ModuleAdapter.hpp
```diff
@@ -107,6 +109,8 @@ struct ModuleAdapter {
     ExtMqttSubscribeFunc ext_mqtt_subscribe;
     std::vector<cmd> registered_commands;
     TelemetryPublishFunc telemetry_publish;
+    GetMappingFunc get_mapping;
+    GetImplMappingFunc get_impl_mapping;
```
a-w50: Is it possible to have only one function instead of two, which then would return a struct with the module mappings and implementation mappings?

a-w50: Btw., we should discuss, if we really need all these `std::functions` here, it might be simpler to have the `ModuleAdapter` just having regular functions and its own private reference to the `everest` instance, and then the functions are implemented by just calling the corresponding functions on the `everest` instance.

include/framework/everest.hpp
```diff
@@ -142,6 +142,16 @@ class Everest {
...
 
+    ///
+    /// \returns the 3 tier model mappings for this module
+    ///
+    std::optional<ModuleTierMappings> get_3_tier_model_mapping();
+
+    ///
+    /// \returns the 3 tier model mapping for the given \p impl_id
+    ///
+    std::optional<Mapping> get_3_tier_model_mapping(const std::string& impl_id);
```
a-w50: In line with the Liskov Substitution Principle, this choice of api naming is maybe a bit unfortunate.  Both methods are named the same, but return different types of objects.  Furthermore they are not `const` correct.
I would advice to only have one accessor method (`const std::optional<ModuleTierMappings>& get_3_tier_model_mapping() const`, note: I would also return by const ref instead of copying).  Additionally you can implement a free helper function like `std::optional<Mapping> get_implementation_mapping(const ModuleTierMapping&, const std::string& impl_id)` which would do the same as `std::optional<Mapping> get_3_tier_model_mapping(const std::string& impl_id)` currently does.  This keeps the interface of the `Everest` class small and also allows you to easily test the function `get_implementation_mapping`.

include/utils/config.hpp
```diff
@@ -147,9 +147,9 @@ class Config {
     json resolve_requirement(const std::string& module_id, const std::string& requirement_id) const;
 
     ///
-    /// \returns a list of Requirements for \p module_id
+    /// \returns a map of Fulfillments for \p module_id
```
a-w50, line 168: I really doubt the value of these comments, you get exactly the same information if you look at the function signature.

include/utils/config.hpp
```diff
@@ -147,9 +147,9 @@ class Config {
...
     ///
-    std::list<Requirement> get_requirements(const std::string& module_id) const;
+    std::map<std::string, std::vector<Fulfillment>> get_fulfillments(const std::string& module_id);
```
a-w50: why isn't this function `const` anymore?

include/utils/types.hpp
```diff
@@ -104,11 +108,41 @@ struct ModuleTierMappings {
...
 struct Requirement {
-    Requirement(const std::string& requirement_id_, size_t index_);
+    Requirement() = default;
+    Requirement(const std::string& requirement_id_, size_t index_, std::optional<Mapping> mapping = std::nullopt);
     bool operator<(const Requirement& rhs) const;
     std::string id;
-    size_t index;
+    size_t index = 0;
+    std::optional<Mapping> mapping;
+};
```
a-w50: Side note: this should just be a simple aggregate.  The `operator<` should be implemented as free function

include/utils/types.hpp
```diff
@@ -104,11 +108,41 @@ struct ModuleTierMappings {
...
+
+/// \brief Describes a how a requirement is fulfilled by a connection to another module and its implementation
```
a-w50: This doc comment should be more like:
A fulfillment relates a requirement to its connected implementation, identified via its module and implementation id.

include/utils/types.hpp
a-w50: I see.  Still I'm unsure, whether the string `Mapping(charging station)` on a text console will help users to understand what is meant.

include/utils/types.hpp
```diff
@@ -104,11 +108,41 @@ struct ModuleTierMappings {
...
+
 struct Requirement {
-    Requirement(const std::string& requirement_id_, size_t index_);
+    Requirement() = default;
+    Requirement(const std::string& requirement_id_, size_t index_, std::optional<Mapping> mapping = std::nullopt);
     bool operator<(const Requirement& rhs) const;
     std::string id;
-    size_t index;
+    size_t index = 0;
+    std::optional<Mapping> mapping;
```
a-w50: I wouldn't add the `mapping` into the `Requirement` struct. The requirement was used as an identifier previously.  Of course it is very convenient, but you could equally add the fulfillments to this struct and so on and so on.  It is probably better to have a separate lookup struct/map/list with Requirement -> Mapping

lib/config.cpp
```diff
@@ -705,23 +705,31 @@ json Config::resolve_requirement(const std::string& module_id, const std::string
     return module_config["connections"][requirement_id];
 }
 
-std::list<Requirement> Config::get_requirements(const std::string& module_id) const {
+std::map<std::string, std::vector<Fulfillment>> Config::get_fulfillments(const std::string& module_id) {
```
a-w50: As mentioned above, I would not add `mapping` to the `Requirement` struct, and hence this function and also the functions using this one, would just stay the same.  Better create the above mentioned struct (which contains the mapping `Requirement/Module -> mapping`) either on the fly or precompute it and then just return that.

lib/config.cpp
```diff
@@ -823,7 +831,8 @@ std::optional<Mapping> Config::get_3_tier_model_mapping(const std::string& modul
     }
     auto& mapping = module_tier_mappings.value();
```
a-w50: This code has been added before, but please use `const` where possible.

lib/config.cpp
```diff
@@ -823,7 +831,8 @@ std::optional<Mapping> Config::get_3_tier_model_mapping(const std::string& modul
...
     if (mapping.implementations.find(impl_id) == mapping.implementations.end()) {
-        return std::nullopt;
+        // if no specific implementation mapping is given, use the module mapping
+        return mapping.module;
     }
     return mapping.implementations.at(impl_id);
```
a-w50, line 870: This pattern happens quite often.  It is usually better to use the iterator instead of doing the lookup twice:
```
const auto mapping_it = ...find(impl_id);
if (mapping_id == std::end(...)) {
  return default;
}
return *mapping_it;
```

lib/everest.cpp
```diff
@@ -158,30 +158,33 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
...
+        for (const Fulfillment& fulfillment : fulfillments.second) {
+            auto& req = fulfillment.requirement;
+            // setup shared database
+            std::shared_ptr<error::ErrorDatabaseMap> error_database = std::make_shared<error::ErrorDatabaseMap>();
+
+            // setup error manager
+            std::string interface_name = this->module_manifest.at("requires").at(req.id).at("interface");
+            json interface_def = this->config.get_interface_definition(interface_name);
+            std::list<std::string> allowed_error_types;
+            for (const auto& error_namespace_it : interface_def["errors"].items()) {
```
a-w50: these changes are probably not necessary, if you keep `get_requirements`

lib/everest.cpp
```diff
@@ -252,6 +255,22 @@ void Everest::register_on_ready_handler(const std::function<void()>& handler) {
...
+std::optional<Mapping> Everest::get_3_tier_model_mapping(const std::string& impl_id) {
+    if (not this->module_tier_mappings.has_value()) {
+        return std::nullopt;
+    }
+    auto& mapping = this->module_tier_mappings.value();
+    if (mapping.implementations.find(impl_id) == mapping.implementations.end()) {
+        // if no specific implementation mapping is given, use the module mapping
+        return mapping.module;
+    }
+    return mapping.implementations.at(impl_id);
```
a-w50: Same story with the iterator as above.  Function should be const and could also return by `const ref&` (for this you would need to have a `static std::optional<Mapping> default_value {std::nullopt}` which you would return, otherwise you directly return the references instead of unecessary copying).

a-w50 (changes requested):
See the comments, please address function signatures and if possible remove the mapping from the `Requirement` struct.

lib/config.cpp
hikinggrass: I guess something like this would be quite readable:
```cpp
const auto& mappings = module_tier_mappings.value();
const auto mapping_it = mappings.implementations.find(impl_id);
if (mapping_it == mappings.implementations.end()) {
    return mappings.module;
}
const auto& [key, mapping] = (*mapping_it);
return mapping;
```

The alternative doesn't quite feel as readable:
```cpp
...
return (*mapping_it).second;
```
Or what do you think?

include/framework/ModuleAdapter.hpp
hikinggrass: Changed into one function, I consider refactoring this out-of-scope for this PR

include/utils/config.hpp
hikinggrass: well spotted, re-added the const

include/utils/types.hpp
hikinggrass: I've applied your suggestion

include/utils/types.hpp
hikinggrass: Removed from Requirement (and added to Fulfillment)

lib/everest.cpp
hikinggrass: I've re-added get_requirements, that does lead to some code duplication though since information required for the Fulfillments is not available from this function

lib/config.cpp
```diff
@@ -710,9 +710,9 @@ std::list<Requirement> Config::get_requirements(const std::string& module_id) co
 
     std::list<Requirement> res;
 
-    std::string module_name = get_module_name(module_id);
+    const std::string module_name = get_module_name(module_id);
```
a-w50: could be `auto`

lib/config.cpp
```diff
@@ -710,9 +710,9 @@ std::list<Requirement> Config::get_requirements(const std::string& module_id) co
...
     for (const std::string& req_id : Config::keys(this->manifests.at(module_name).at("requires"))) {
-        json resolved_req = this->resolve_requirement(module_id, req_id);
+        const json resolved_req = this->resolve_requirement(module_id, req_id);
```
a-w50: could be `auto`

include/utils/types.hpp
```diff
@@ -104,11 +108,42 @@ struct ModuleTierMappings {
...
+    size_t index = 0;
+};
+
+/// \brief A Fulfillment relates a Requirement to its connected implementation, identified via its module and
+/// implementation id. It additionally provides access to the Mapping associated with the Requirement.
+struct Fulfillment {
+    std::string module_id;
+    std::string implementation_id;
+    Requirement requirement;
+    std::optional<Mapping> mapping;
```
a-w50: Looking at the `everestpy` code, a `Fulfillment` only contains the module id and implementation id.  I should probably have named that `FulfillmentID` back then (and similar `RequirementID`).
What I've said about putting the mapping into the `Requirement` in the previous review, also applies to the `Fulfillment` here.  `Fulfillment` and `Requirement` should only be identifiers.  Their connection should be represented as `Requirement -> Fulfillment` (as is done in everestpy).
What should be done here, is to have a function like `resolve_requirements(module id)` which returns a map (or a vector) of type `std::map<Requirement, Fulfillment>` (this is actually the logic/code shared by `get_requirements` and `get_fulfillments` - both functions should use this common function then).
Having this map, you would then need to transform it:
- in case of `get_requirements` you only need to return the `keys`
- in case of `get_fulfillments`, this should be the identity, i.e. just returning the connections

Now you still don't have access to the mapping.  In order to expose that, you'll need to do a transform of the above mentioned `std::map`, where you iterate over the keys (requirements) and transform the fulfillments into their corresponding mapping, i.e. a map like `std::map<Requirement, Mapping>`.  And with this you should proceed then. This function could be called `transform_to_mapping`, ideally a free function which takes a functor (std::function), which can retrieve a mapping from a fulfillment.

The basic idea behind this separation is to have identifier structs (like Requirement or Fulfillment).  Now we may have different parts in the code which want to attach information to these identifiers (like attaching a Fulfillment to a Requirement, or attaching a Mapping to a Fulfillment or Requirement).  If we would put all that stuff into the Requirement or Fulfillment identifiers, we create coupling between these sections, and we should avoid that.  We achieve better separation, when we create new or custom types (like an additional map), which each contain specific mappings.

lib/runtime.cpp
```diff
@@ -495,11 +497,12 @@ int ModuleLoader::initialize() {
...
         };
 
+        module_adapter.get_mapping = [&everest]() { return everest.get_3_tier_model_mapping(); };
+
         this->callbacks.register_module_adapter(module_adapter);
 
         // FIXME (aw): would be nice to move this config related thing toward the module_init function
-        std::vector<cmd> cmds =
-            this->callbacks.everest_register(config.get_main_config()[this->module_id]["connections"]);
+        std::vector<cmd> cmds = this->callbacks.everest_register(config.get_fulfillments(this->module_id));
```
a-w50: I just saw that now, why do we pass the fulfillments via `everest_register`?  `everest_register` wasn't intended to transfer this kind of information to the module - it's intend was to get the implemented commands of the module (if I remember correctly).
This information should be passed via the init function.  If possible, I would add this information to the `ModuleInfo` struct, in case it can be consumed directly via user code, or if not pass it as an additional argument and then it's consumption needs to take place in the `ld-ev` code which finds the corresponding requirement and sets its `_mapping` from that.

a-w50 (changes requested):

include/utils/types.hpp
```diff
@@ -132,20 +132,29 @@ struct TelemetryConfig {
...
 
+bool operator<(const Requirement& lhs, const Requirement& rhs);
+
 /// \brief A Fulfillment relates a Requirement to its connected implementation, identified via its module and
-/// implementation id. It additionally provides access to the Mapping associated with the Requirement.
+/// implementation id.
 struct Fulfillment {
     std::string module_id;
     std::string implementation_id;
     Requirement requirement;
```
a-w50, line 144: Is there any reason, why requirement is still here?

include/utils/types.hpp
```diff
@@ -132,20 +132,29 @@ struct TelemetryConfig {
...
+};
+
+/// \brief Contains everything that's needed to initialize a requirement in user code
+struct RequirementInitializer {
+    Requirement requirement;
+    Fulfillment fulfillment;
     std::optional<Mapping> mapping;
 };
 
+using RequirementInitialization = std::map<std::string, std::vector<RequirementInitializer>>;
```
a-w50, line 154: I'm uncertain with that name:  `RequirementInitialization` sounds more like the result of the initialization of a requirement than a mapping from each requirement name to its initializer.

lib/config.cpp
```diff
@@ -705,25 +705,40 @@ json Config::resolve_requirement(const std::string& module_id, const std::string
...
-    std::list<Requirement> res;
-
-    const std::string module_name = get_module_name(module_id);
-    for (const std::string& req_id : Config::keys(this->manifests.at(module_name).at("requires"))) {
+    const auto module_name = get_module_name(module_id);
+    for (const auto& req_id : Config::keys(this->manifests.at(module_name).at("requires"))) {
         const json resolved_req = this->resolve_requirement(module_id, req_id);
         if (!resolved_req.is_array()) {
+            const auto resolved_module_id = resolved_req.at("module_id");
+            const auto resolved_impl_id = resolved_req.at("implementation_id");
```
a-w50: These should be `auto&`, right?

lib/config.cpp
```diff
@@ -705,25 +705,40 @@ json Config::resolve_requirement(const std::string& module_id, const std::string
...
+}
+
+std::list<Requirement> Config::get_requirements(const std::string& module_id) const {
+    BOOST_LOG_FUNCTION();
+
+    std::list<Requirement> res;
+
+    for (const auto& [requirement, fulfillment] : this->resolve_requirements(module_id)) {
+        res.push_back(requirement);
+        (void)fulfillment; // fulfillment is not used here
```
a-w50: This void is a not that beautiful.  `[[maybe_unused]]` could be used?  Or just normal `it->first`?

lib/config.cpp
```diff
@@ -732,26 +747,21 @@ std::map<std::string, std::vector<Fulfillment>> Config::get_fulfillments(const s
...
-        }
-        res[req_id] = fulfillments;
+    for (const auto& [requirement, fulfillment] : this->resolve_requirements(module_id)) {
+        res[requirement.id].push_back(fulfillment);
+    }
+
+    return res;
+}
+
+RequirementInitialization Config::get_requirement_initialization(const std::string& module_id) const {
```
a-w50, line 757: This function is now _very_ readable :+1:

lib/config.cpp
```diff
@@ -732,26 +747,21 @@ std::map<std::string, std::vector<Fulfillment>> Config::get_fulfillments(const s
...
+    BOOST_LOG_FUNCTION();
+
+    RequirementInitialization res;
+
+    for (const auto& [requirement, fulfillment] : this->resolve_requirements(module_id)) {
+        const auto& mapping = this->get_3_tier_model_mapping(fulfillment.module_id, fulfillment.implementation_id);
+        res[requirement.id].push_back({requirement, fulfillment, mapping});
```
a-w50, line 764: This is a bit tricky. We implicitly rely on the fact, that requirements are sorted within the map according to their index.  This should be guaranteed, but still is not obvious to the reader.  Having the requirement without it's index, but instead pointing at a vector of `Fulfillments` would probably make the code even simpler.

include/utils/types.hpp
```diff
@@ -132,20 +132,29 @@ struct TelemetryConfig {
 struct Requirement {
     Requirement() = default;
     Requirement(const std::string& requirement_id_, size_t index_);
```
a-w50: These could be dropped now

include/utils/types.hpp
hikinggrass: It's actively used in everestpy and I haven't looked into how to replace it yet, might do that though

include/utils/types.hpp
hikinggrass: I'm open for suggestions :smile: that was the best I could come up with at the time :sweat_smile:

include/utils/types.hpp
hikinggrass: Ah I thought I removed them already, must have overlooked that

include/utils/types.hpp
a-w50: I see, sorry I wasn't aware of that

a-w50 (approved):

PR #208, opened 2024-10-10
Fix: MQTTAbstractionImpl: flush/sync MQTT Connect message immediately

barsnick:
Previously, a socket was opened, but the MQTT Connect command message was not flushed to the broker until init() was finished. This could be critical if a module's init() phase waits for some reason, as the MQTT broker's default socket timeout might disconnect the client.

This behavior was observed with the mosquitto broker, where unregistered sockets were consistently disconnected after 94 seconds.

Insert an additional mqtt_sync() directly after mqtt_connect(), while the mqtt_sync() thread is not yet started.

Co-authored-by: Fabian Hartung <fabian.hartung@chargebyte.com>

### DIFF
--- a/lib/mqtt_abstraction_impl.cpp
+++ b/lib/mqtt_abstraction_impl.cpp
@@ -484,19 +484,18 @@ bool MQTTAbstractionImpl::connectBroker(std::string& socket_path) {
     uint8_t connect_flags = MQTT_CONNECT_CLEAN_SESSION;
     /* Send connection request to the broker. */
     if (mqtt_connect(&this->mqtt_client, nullptr, nullptr, nullptr, 0, nullptr, nullptr, connect_flags,
-                     mqtt_keep_alive) == MQTT_OK) {
-        // TODO(kai): async?
-        MQTTErrors error = mqtt_sync(&this->mqtt_client);
-        if (error != MQTT_OK) {
-            EVLOG_error << fmt::format("Error during MQTT sync: {}", mqtt_error_str(error));
-            return false;
-        }
-
-        on_mqtt_connect();
-        return true;
+                     mqtt_keep_alive) != MQTT_OK) {
+        return false;
+    }
+    // TODO(kai): async?
+    const auto error = mqtt_sync(&this->mqtt_client);
+    if (error != MQTT_OK) {
+        EVLOG_error << fmt::format("Error during MQTT sync: {}", mqtt_error_str(error));
+        return false;
     }
 
-    return false;
+    on_mqtt_connect();
+    return true;
 }
 
 bool MQTTAbstractionImpl::connectBroker(const char* host, const char* port) {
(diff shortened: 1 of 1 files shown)


mhei:
Is calling the `on_mqtt_disconnect()` callback reasonable in the error path at this point? I'm wondering about the API contract regarding the callbacks call sequence, i.e. I would expect that `on_mqtt_disconnect()` is only called once `on_mqtt_connect()` was called before.
Disclaimer: I did not dig deep into the whole code base so this comment might be stupid.

FaHaGit:
We also stumbled across the point of whether we should call `on_mqtt_disconnect()`. Now the same error handling is implemented as in the `spawn_main_loop_thread()` method, so it should behave the same as before. But I admit a disconnect without a connect first is a bit confusing.

FaHaGit:
I have now removed the disconnect() call, as no disconnection should occur unless an on_connect() call has been made. We also considered calling spawn_main_loop_thread() before the callback of the init() method, but that would be a bigger change and may influence the current behaviour. We see a higher risk here that MQTT messages could be lost if a module publishes messages within the init() method.

mhei:
LGTM now

hikinggrass (approved):

lib/mqtt_abstraction_impl.cpp
```diff
@@ -486,6 +486,12 @@ bool MQTTAbstractionImpl::connectBroker(std::string& socket_path) {
     if (mqtt_connect(&this->mqtt_client, nullptr, nullptr, nullptr, 0, nullptr, nullptr, connect_flags,
                      mqtt_keep_alive) == MQTT_OK) {
         // TODO(kai): async?
+        MQTTErrors error = mqtt_sync(&this->mqtt_client);
+        if (error != MQTT_OK) {
+            EVLOG_error << fmt::format("Error during MQTT sync: {}", mqtt_error_str(error));
+            return false;
+        }
+
```
a-w50: The original code wasn't using early returns, but here might be a good place to change that.  Something like:
```
if (mqtt_connect(..) != MQTT_OK) {
  return false;
}

if ((const auto error = mqtt_sync(...)) != MQTT_OK) {
  EVLOG...;
  return false;
}

on_mqtt_connect();
return true
```

More readable imho.

As was mentioned in the comments already, a possible better solution would be to start the _event loop_/_the polling thread_ directly after doing the connect - because why would you create a socket and then not use it for some time?  It is indeed interesting to see that the mqttc implementation doesn't send anything on the `mqtt_connect` command - probably due to easier separation for error handling.  Good find @FaHaGit

a-w50:
> I have now removed the disconnect() call, as no disconnection should occur unless an on_connect() call has been made. We also considered calling spawn_main_loop_thread() before the callback of the init() method, but that would be a bigger change and may influence the current behaviour. We see a higher risk here that MQTT messages could be lost if a module publishes messages within the init() method.

Don't be afraid of bigger changes, otherwise they grow even bigger :)

FaHaGit:
Thank you for the feedback @a-w50! I have adjusted the code accordingly. If it is ok we would stay with the initial solution for now. We can also prepare a separate pull request to try out the other (more cleaner) solution.

a-w50:
> Thank you for the feedback @a-w50! I have adjusted the code accordingly. If it is ok we would stay with the initial solution for now. We can also prepare a separate pull request to try out the other (more cleaner) solution.


I think it is ok for now.  Feel free to create an issue for the follow up work.  Both `connectBroker` function overloads should be refactored anyway, as they contain quite some code duplication.

FaHaGit:
I have created the [issues 213](https://github.com/EVerest/everest-core/issues/1618) regarding the open points. Should I squash & merge the pull request now, or will you take over when the time is right regarding the release process? Thank you

PR #210, opened 2024-10-14
Add explicit exceptions during command calls / handling

hikinggrass:
Allow exceptions thrown in cmd handlers to be propagated to the calling module.

To enable this you can set the following in a config file:
```yaml
settings:
  forward_exceptions: true
````

This allows the caller to handle certain error conditions more gracefully (and could be extended for shutdown handling during cmd callc in the future as well)

A module could handle these new exceptions in the following way:

```cpp
try {
    req->call_something();
} catch (const Everest::HandlerException& e) {
    EVLOG_error << "got a HandlerException: " << e.what();
} catch (const Everest::CmdTimeout& e) {
    EVLOG_error << "got a CmdTimeout: " << e.what();
} catch (const Everest::CmdError& e) {
    EVLOG_error << "got a CmdError: " << e.what();
}
```

`HandlerExceptions` and `CmdTimeout` inherit from `CmdError` so a caller could also just catch this.


include/framework/everest.hpp
```diff
@@ -34,6 +34,49 @@ using TelemetryEntry = std::variant<std::string, const char*, bool, int32_t, uin
 using TelemetryMap = std::map<std::string, TelemetryEntry>;
 using UnsubscribeToken = std::function<void()>;
 
+enum class ErrorType {
```
hikinggrass: `ErrorType` might be a bit vague and could lead to confusions with the "error framework"

include/framework/everest.hpp
```diff
@@ -34,6 +34,49 @@ using TelemetryEntry = std::variant<std::string, const char*, bool, int32_t, uin
...
+    MessageParsing,
+    SchemaValidation,
+    HandlerException,
+    Timeout,
+    Shutdown,
+    Unknown
```
hikinggrass: remove Unknown

include/framework/everest.hpp
hikinggrass: Maybe call this `CallCmdEvent` or `CmdEvent`

include/framework/everest.hpp
```diff
@@ -34,6 +34,49 @@ using TelemetryEntry = std::variant<std::string, const char*, bool, int32_t, uin
...
+};
+
+struct ErrorMessage {
+    ErrorType type;
```
hikinggrass: event

include/framework/everest.hpp
```diff
@@ -34,6 +34,49 @@ using TelemetryEntry = std::variant<std::string, const char*, bool, int32_t, uin
...
+enum class ErrorType {
+    MessageParsing,
+    SchemaValidation,
+    HandlerException,
+    Timeout,
+    Shutdown,
+    Unknown
+};
+
+struct ErrorMessage {
```
hikinggrass: `CmdResultError`

include/framework/everest.hpp
```diff
@@ -34,6 +34,49 @@ using TelemetryEntry = std::variant<std::string, const char*, bool, int32_t, uin
...
+class HandlerException : public CmdError {
+public:
+    using CmdError::CmdError;
+};
+
+class CmdTimeout : public CmdError {
+public:
+    using CmdError::CmdError;
+};
+
```
hikinggrass: Maybe already define exceptions for the other `CallCmdEvent`s

lib/everest.cpp
```diff
@@ -863,19 +895,26 @@ void Everest::provide_cmd(const std::string& impl_id, const std::string& cmd_nam
...
-        res_data["retval"] = handler(data.at("args"));
+        try {
+            if (not error.has_value()) {
+                res_data["retval"] = handler(data.at("args"));
+            }
+        } catch (const std::exception& e) {
+            EVLOG_verbose << fmt::format("Exception during handling of: {}->{}({}): {}",
+                                         this->config.printable_identifier(this->module_id, impl_id), cmd_name,
+                                         fmt::join(arg_names, ","), e.what());
+            error = ErrorMessage{ErrorType::HandlerException, e.what()};
```
hikinggrass: re-throw after result has been sent to caller so the handler still crashes

hikinggrass:
TODO: add test-cases

lib/everest.cpp
```diff
@@ -879,16 +923,30 @@ void Everest::provide_cmd(const std::string& impl_id, const std::string& cmd_nam
...
+        if (error.has_value()) {
+            auto err = error.value();
+            if (err.event == CmdEvent::HandlerException) {
+                if (err.ex) {
+                    std::rethrow_exception(err.ex);
+                } else {
+                    throw std::runtime_error("Unknown exception during cmd handling");
+                }
+            }
+        }
```
hikinggrass: TODO: should this be configurable? default = rethrow (which is very similar to the current behavior) but can be turned off?

include/utils/types.hpp
```diff
@@ -187,9 +187,59 @@ NLOHMANN_JSON_NAMESPACE_END
 #define EVCALLBACK(function) [](auto&& PH1) { function(std::forward<decltype(PH1)>(PH1)); }
 
 namespace Everest {
+
+enum class CmdEvent {
+    MessageParsingFailed,
+    SchemaValidationFailed,
+    HandlerException,
+    Timeout,
+    Shutdown
```
dorezyuk: should we add also a notyetready?

lib/conversions.cpp
```diff
@@ -0,0 +1,61 @@
...
+constexpr auto CMD_EVENT_SCHEMA_VALIDATION_FAILED = "SchemaValidationFailed";
+constexpr auto CMD_EVENT_HANDLER_EXCEPTION = "HandlerException";
+constexpr auto CMD_EVENT_TIMEOUT = "Timeout";
+constexpr auto CMD_EVENT_SHUTDOWN = "Shutdown";
+
+std::string cmd_event_to_string(CmdEvent cmd_event) {
+    switch (cmd_event) {
+    case CmdEvent::MessageParsingFailed:
+        return CMD_EVENT_MESSAGE_PARSING_FAILED;
+        break;
```
Pietfried: Can be removed here and following
```suggestion
```

lib/everest.cpp
```diff
@@ -401,18 +411,32 @@ json Everest::call_cmd(const Requirement& req, const std::string& cmd_name, json
...
+            throw HandlerException(fmt::format("{}", error.msg));
+        } else if (error.event == CmdEvent::Timeout) {
+            throw CmdTimeout(fmt::format("{}", error.msg));
+        }
+        throw CmdError(fmt::format("{}: {}", conversions::cmd_event_to_string(error.event), error.msg));
+    } else if (not result.result.has_value()) {
+        throw CmdError("Command did not return result");
+    } else {
+        return result.result.value();
+    }
```
Pietfried: nit: reads a little better
```suggestion
if (result.error.has_value()) {
        const auto& error = result.error.value();
        const std::string err_msg = fmt::format("{}", error.msg);
    
        switch (error.event) {
            case CmdEvent::HandlerException:
                throw HandlerException(err_msg);
            case CmdEvent::Timeout:
                throw CmdTimeout(err_msg);
            default:
                throw CmdError(fmt::format("{}: {}", conversions::cmd_event_to_string(error.event), error.msg));
        }
    }
    
    if (!result.result.has_value()) {
        throw CmdError("Command did not return result");
    }
    
    return result.result.value();
```

lib/everest.cpp
```diff
@@ -851,19 +883,31 @@ void Everest::provide_cmd(const std::string& impl_id, const std::string& cmd_nam
...
+        } catch (const std::exception& e) {
+            EVLOG_error << fmt::format("Exception during handling of: {}->{}({}): {}",
+                                       this->config.printable_identifier(this->module_id, impl_id), cmd_name,
+                                       fmt::join(arg_names, ","), e.what());
+            error = CmdResultError{CmdEvent::HandlerException, e.what(), std::current_exception()};
+        } catch (...) {
+            EVLOG_error << fmt::format("Unknown exception during handling of: {}->{}({})",
+                                       this->config.printable_identifier(this->module_id, impl_id), cmd_name,
+                                       fmt::join(arg_names, ","));
+            error = CmdResultError{CmdEvent::HandlerException, "Unknown exception"};
```
Pietfried: couldnt we use `std::current_exception()` here as well?

lib/everest.cpp
Pietfried: For now I would just rethrow it and possibly make it configurable in a subsequent PR after we have discussed and analyzed potential implications of that

include/utils/types.hpp
```diff
@@ -187,9 +187,59 @@ NLOHMANN_JSON_NAMESPACE_END
...
+
+struct CmdResultError {
+    CmdEvent event;
+    std::string msg;
+    std::exception_ptr ex;
+};
+
+struct CmdResult {
+    std::optional<json> result;
+    std::optional<CmdResultError> error;
```
dorezyuk, line 154: to my understanding they are mutually exclusive - can we express that through the type system, e.x a variant?

include/utils/types.hpp
hikinggrass: added NotReady enum

include/utils/types.hpp
hikinggrass: A variant here might a bit tricky at the moment due to `CmdResultError` being convertible to `json` and result being a `json` type. But let's keep this in mind during the upcoming refactorings in this part of the code

lib/everest.cpp
hikinggrass: added usage of `std::current_exception()`

lib/everest.cpp
```diff
@@ -401,18 +411,37 @@ json Everest::call_cmd(const Requirement& req, const std::string& cmd_name, json
...
-    return result;
+    if (result.error.has_value()) {
+        const auto& error = result.error.value();
+        const auto error_message = fmt::format("{}", error.msg);
+        switch (error.event) {
+        case CmdEvent::HandlerException:
+            throw HandlerException(error_message);
+        case CmdEvent::Timeout:
+            throw CmdTimeout(error_message);
+        default:
```
hikinggrass, line 444: TODO: use other defined exceptions here

include/utils/types.hpp
```diff
@@ -187,9 +187,66 @@ NLOHMANN_JSON_NAMESPACE_END
...
+};
+
 struct BootException : public std::runtime_error {
     using std::runtime_error::runtime_error;
 };
+
+class CmdError : public std::runtime_error {
```
hikinggrass: move this into a separate file

lib/everest.cpp
hikinggrass: default is now still the current behavior, but the new behavior can be turned on by setting the following in a config:
```yaml
settings:
  forward_exceptions: true
````

lib/everest.cpp
Pietfried: This doesnt look like it is resolved?

What about?
MessageParsingFailed
SchemaValidationFailed
Shutdown
NotReady

Pietfried (changes requested):
lgtm, except one comment in line

lib/everest.cpp
hikinggrass: Good catch, I have added the missing exception throws

Pietfried (approved):

PR #211, opened 2024-10-18
Date: adjust flags to not use remote

dorezyuk:
Fixes the compilation flags for the date library 
The flags were not the same as for the cmake/yocto toolchains


hikinggrass:
>[...] The flags are not the same as for the cmake/yocto toolchains

Why are they different?

dorezyuk:
> Why are they different?

it should be past tense - they were different and this commit fixes that

hikinggrass (approved):

PR #212, opened 2024-10-23
Rust: Add support for the errors

### LINKED ISSUE #175: Support Everest Errors Raising and handling in Rust
Everest Framework supports errors that can be raised and handled by modules.

Rust Implementation is missing this feature. 
Side note: We may need to wait until the concept of errors will stabilize, since it is still work in progress.

dorezyuk:
Adds error support for rust. 
Closes #175

### DIFF
--- a/everestrs/everestrs-build/jinja/errors.jinja2
+++ b/everestrs/everestrs-build/jinja/errors.jinja2
@@ -1,30 +1,27 @@
 {%- for name, errors in involved_errors | items  %}
-    mod {{ name | snake }} {
-
-    {%- for error in errors %}
-
+    pub mod {{ name | snake }} {
+        use everestrs::serde as __serde;
+    {%- for error_group in errors %}
         /// The error definition of the {{ name }} interface.
-        /// {{ error.description | replace("\n", " ") }}
+        /// {{ error_group.error_list.description | replace("\n", " ") }}
         #[derive(Debug, Clone, __serde::Serialize, __serde::Deserialize)]
         #[serde(crate = "__serde")]
-        enum {{ error.name | title }}Error {
-            {%- for error_entry in error.errors %}
+        pub enum {{ error_group.name | title }}Error {
+            {%- for error_entry in error_group.error_list.errors %}
             /// {{ error_entry.description | replace("\n", " ")}}
             #[serde(rename = "{{ name }}/{{ error_entry.name }}")]
             {{ error_entry.name | title}},
             {%- endfor %}
         }
-
     {%- endfor %}
-
         /// All possible errors of the {{ name }} interface.
         #[derive(Debug, Clone, __serde::Serialize, __serde::Deserialize)]
         #[serde(crate = "__serde")]
-        enum {{name | title}}Error {
-            {%- for error_entry in error.errors %}
+        pub enum Error {
+    {%- for error_group in errors %}
             #[serde(flatten)]
-            {{ error_entry.name | title}}({{ error_entry.name | title}}),
-            {%- endfor %}
+            {{ error_group.name | title }}Error({{ error_group.name | title }}Error),
+    {%- endfor %}
         }
     }
 {%- endfor %}
--- a/everestrs/everestrs-build/jinja/errors.jinja2
+++ b/everestrs/everestrs-build/jinja/errors.jinja2
@@ -20,7 +20,7 @@
         #[serde(untagged)]
         pub enum Error {
     {%- for error_group in errors %}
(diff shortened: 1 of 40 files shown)


everestrs/everestrs-build/jinja/errors.jinja2
```diff
@@ -0,0 +1,27 @@
+{%- for name, errors in involved_errors | items  %}
+    pub mod {{ name | snake }} {
```
ju4nlu, line 2: Seems like all generated lines from this file come with an extra indentation, isn't it the case?

everestrs/everestrs-build/jinja/module.jinja2
```diff
@@ -211,6 +216,32 @@ impl ::everestrs::Subscriber for Module {
...
+{% for req in requires %}
+            "{{ req.implementation_id }}" => {
+    {% if req.min_connections == 1 and req.max_connections == 1 %}
+                dispatch_error_to_{{ req.interface | snake }}(&context, self.{{ req.implementation_id  | identifier }}.as_ref(), error, raised)
+    {% else %}
+                dispatch_error_to_{{ req.interface | snake }}(&context, self.{{ req.implementation_id  | identifier }}_slots[index].as_ref(), error, raised)
+    {% endif %}
+            },
+{% endfor %}
+            _ => eprintln!("Received an unknonw error from {implementation_id}"),
```
ju4nlu: typo
```suggestion
            _ => eprintln!("Received an unknown error from {implementation_id}"),
```

everestrs/everestrs-build/jinja/module.jinja2
```diff
@@ -211,6 +216,32 @@ impl ::everestrs::Subscriber for Module {
...
+        error: ::everestrs::ErrorType,
+        raised: bool
+    ) {
+        let context = Context {
+            publisher: &self.publisher,
+            name: implementation_id,
+            index,
+        };
+        match implementation_id {
+{% for req in requires %}
```
ju4nlu, line 232: indenting the jinja lines at the same level as the rust code will improve the reading experience

everestrs/everestrs-build/src/codegen.rs
```diff
@@ -307,12 +315,145 @@ impl CommandContext {
...
+            let path = splits.next().ok_or(anyhow::anyhow!("No path defined"))?;
+
+            // Split the path and remove the empty parts.
+            // (The first element might be empty if we have a leading `/`).
+            let paths = path
+                .split("/")
+                .filter(|path| !path.is_empty())
+                .collect::<Vec<_>>();
+
+            anyhow::ensure!(paths.len() == 2);
```
ju4nlu: Is the message printed if this is not fulfilled meaningful enough? I believe it will just print `Error: Condition failed: `paths.len() == 2` (1 vs 2)`. Same happens with the line below this one

everestrs/everestrs-build/src/codegen.rs
```diff
@@ -651,3 +811,46 @@ pub fn emit(manifest_path: PathBuf, everest_core: Vec<PathBuf>) -> Result<String
...
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_split_paths_invalid() {
+        use super::impl_error::*;
+        let invalid_input = [
+            "/foo/bar/baz", // too many
+            "/foo",         // too view,
```
ju4nlu:
```suggestion
            "/foo",         // too few,
```

I guess

schemas/error-declaration-list.yaml
```diff
@@ -21,10 +21,6 @@ $defs:
       description:
         type: string
         minLength: 2
-      namespace:
-        type: string
-        minLength: 2
-        pattern: ^[a-z][a-z0-9]*$
```
ju4nlu, line 27: Not really related to supporting errors in EVerest, I think this should be a part of a different MR

dorezyuk: I've discussed this with @andistorm - this just updates the schema

everestrs/everestrs-build/jinja/errors.jinja2
dorezyuk: these lines are include in the errors module in the the module.jinja file - so they must be indented

everestrs/everestrs-build/jinja/module.jinja2
dorezyuk: we did it like this mostly through out the file indenting rust and jinda independently

SirVer (approved):

PR #214, opened 2024-10-28
Use shared error type map

dorezyuk:
Currently the error type map is copy constructed and then placed into a shared pointer which results in the error map not being really shared between the different classes. This Pr fixes it.

### DIFF
--- a/include/utils/error/error_factory.hpp
+++ b/include/utils/error/error_factory.hpp
@@ -16,10 +16,11 @@ namespace error {
 
 class ErrorFactory {
 public:
-    explicit ErrorFactory(ErrorTypeMapPtr error_type_map);
-    ErrorFactory(ErrorTypeMapPtr error_type_map, ImplementationIdentifier default_origin);
-    ErrorFactory(ErrorTypeMapPtr error_type_map, ImplementationIdentifier default_origin, Severity default_severity);
-    ErrorFactory(ErrorTypeMapPtr error_type_map, std::optional<ImplementationIdentifier> default_origin,
+    explicit ErrorFactory(ErrorTypeMap::ConstPtr error_type_map);
+    ErrorFactory(ErrorTypeMap::ConstPtr error_type_map, ImplementationIdentifier default_origin);
+    ErrorFactory(ErrorTypeMap::ConstPtr error_type_map, ImplementationIdentifier default_origin,
+                 Severity default_severity);
+    ErrorFactory(ErrorTypeMap::ConstPtr error_type_map, std::optional<ImplementationIdentifier> default_origin,
                  std::optional<Severity> default_severity, std::optional<State> default_state,
                  std::optional<ErrorType> default_type, std::optional<ErrorSubType> default_sub_type,
                  std::optional<std::string> default_message, std::optional<std::string> default_vendor_id);
@@ -50,7 +51,7 @@ private:
     std::optional<std::string> default_message;
     std::optional<std::string> default_vendor_id;
 
-    ErrorTypeMapPtr error_type_map;
+    ErrorTypeMap::ConstPtr error_type_map;
 
     void set_description(Error& error) const;
 };
--- a/include/utils/error/error_factory.hpp
+++ b/include/utils/error/error_factory.hpp
@@ -9,19 +9,17 @@
 #include <string>
 
 #include <utils/error.hpp>
+#include <utils/error/error_type_map.hpp>
 
 namespace Everest {
 namespace error {
 
-struct ErrorTypeMap;
-
 class ErrorFactory {
 public:
-    explicit ErrorFactory(std::shared_ptr<ErrorTypeMap> error_type_map);
-    ErrorFactory(std::shared_ptr<ErrorTypeMap> error_type_map, ImplementationIdentifier default_origin);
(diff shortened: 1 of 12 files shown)


include/utils/error/error_factory.hpp
```diff
@@ -52,7 +51,7 @@ class ErrorFactory {
     std::optional<std::string> default_message;
     std::optional<std::string> default_vendor_id;
 
-    const std::shared_ptr<ErrorTypeMap> error_type_map;
```
dorezyuk, line 55: I suppose the intention was to say a shared pointer to a const ErrorMap and not a const shared pointer to a (mutable) ErrorMap

include/utils/error/error_manager_req.hpp
```diff
@@ -14,13 +15,12 @@ namespace Everest {
 namespace error {
 
 struct ErrorDatabase;
-struct ErrorTypeMap;
```
dorezyuk, line 17: I don't really get what the idea behind all the forward defines was - I've removed it to just use the ErrorTypeMapPtr type

include/utils/error/error_type_map.hpp
```diff
@@ -50,6 +53,10 @@ class ErrorTypeMap {
     ///
     bool has(const ErrorType& error_type) const;
 
+    /// Const pointer to a ErrorTypeMap. This is the default how we share the
+    /// error type map between different classes.
+    using ConstPtr = std::shared_ptr<const ErrorTypeMap>;
+
```
andistorm, line 59: I would prefer to keep using `std::shared_ptr<const ErrorTypeMap>` since it fits the style arround here

include/utils/error/error_type_map.hpp
```diff
@@ -5,6 +5,9 @@
 #define UTILS_ERROR_TYPE_MAP_HPP
 
 #include <filesystem>
+#include <map>
+#include <memory>
```
andistorm, line 9:
```suggestion
```
This dependency can be avoided, when not defining this `ConstPtr`

include/utils/error/error_factory.hpp
```diff
@@ -9,19 +9,18 @@
 #include <string>
 
 #include <utils/error.hpp>
+#include <utils/error/error_type_map.hpp>
 
 namespace Everest {
 namespace error {
 
-struct ErrorTypeMap;
```
andistorm, line 16: I would like to keep this forward declaration to avoid unnecessary dependencies in all these header files

include/utils/error/error_manager_req.hpp
andistorm: The idea of forward declarations is to avoid including dependencies in the header file to reduce compile times after changing something in the included files

lib/config.cpp
```diff
@@ -397,7 +397,7 @@ std::tuple<json, int> Config::load_and_validate_with_schema(const fs::path& file
 Config::Config(std::shared_ptr<RuntimeSettings> rs) : Config(rs, false) {
 }
 
-Config::Config(std::shared_ptr<RuntimeSettings> rs, bool manager) : rs(rs), manager(manager) {
+Config::Config(std::shared_ptr<RuntimeSettings> rs, bool manager) : rs(rs), manager(manager), error_map{nullptr} {
```
andistorm, line 400: This default value is not needed, especially because it is initialized in the body of the constructor

lib/everest.cpp
```diff
@@ -103,9 +102,9 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
         error::ErrorManagerImpl::PublishErrorFunc publish_cleared_error = [this, impl](const error::Error& error) {
             this->publish_cleared_error(impl, error);
         };
-        this->impl_error_managers[impl] = std::make_shared<error::ErrorManagerImpl>(
-            std::make_shared<error::ErrorTypeMap>(this->config.get_error_map()), error_database, allowed_error_types,
-            publish_raised_error, publish_cleared_error);
+        this->impl_error_managers[impl] =
+            std::make_shared<error::ErrorManagerImpl>(this->config.get_error_map(), error_database, allowed_error_types,
+                                                      publish_raised_error, publish_cleared_error);
```
andistorm, line 107: I don't really get what changed here

andistorm (changes requested):
Thanks for your work here @dorezyuk,

I really would like to merge the optimization of sharing the pointer instead of copying the ErrorTypeMap.
Could you please move the style changes in another PR/Discussion?
Regarding the forward declarations I see a disadvantage in removing them since they can safe compile time by reducing dependencies between the header files.

Please let me know if you have further questions to my comments

include/utils/error/error_type_map.hpp
dorezyuk: not sure if i understand - we're using it

include/utils/error/error_type_map.hpp
dorezyuk: also not sure if I understand - we use it in this library a lot. What do you mean with "avoid" the dependency? Are you concerned about the preprocessor adding this?

andistorm:
I’m closing this PR due to prolonged inactivity, unresolved comments, and the fact that the changes are somewhat outdated. Another important reason is the upcoming restructuring of the code, as this repository will be integrated into the monorepo. Given these factors, it doesn’t make sense to move this PR forward at this time.

@dorezyuk , thank you for the contribution and discussion so far.

PR #218, opened 2024-11-11
Ensure that "lib" is the CMAKE_INSTALL_LIBDIR in some bazel dependencies

hikinggrass:
If this is not set on some systems this might default to lib64 leading to bazel build errors like "...libcurl.a' was not created


hikinggrass:
An alternative would be to properly support this in the bazel build

dorezyuk:
Hi I can’t verify it right now since I’m currently traveling…
But I’m wondering how you build it and why this happens. We compile this for x86 and aarch64 targets and didn’t run into this issue. If it’s urgent feel free to merge otherwise we can have a look together once I’m back:)

hikinggrass:
> Hi I can’t verify it right now since I’m currently traveling… But I’m wondering how you build it and why this happens. We compile this for x86 and aarch64 targets and didn’t run into this issue. If it’s urgent feel free to merge otherwise we can have a look together once I’m back:)

Not urgent at all, enjoy your vacation :palm_tree: 
I'm building with bazelisk on opensuse tumbleweed, there these artifacts end up in lib64 while bazel seems to expect them in lib

dorezyuk (approved):

andistorm (approved):
Approve since @dorezyuk approved but hasn't write access

PR #224, opened 2024-12-21
Various performance improvements

hikinggrass:
* contains performance enhancements in:
  * schema loading, config parsing and validation
  * async MQTT config distribution
  * publishing of manifests and module names
  * handling of types of config items where until now the whole schema was published but only the type was needed

* Additionally addresses some clang-tidy issues

* Adds a new ctor for everestpy to allow MQTT based config distribution to be used in everest-testing more easily

* Remove retained topics after startup and add flag that prevents this from happening for introspection

Some quick-and-dirty performance comparisons starting up the config-sil with a recent everest-core main. This mostly shows that config loading got faster (by cleaning up the usage of schema validators) and a decrease in individual module startup time. Interestingly the total startup time is pretty similar which warrants further investigation:

(v0.19.2 vs v0.19.1)

```
manager          :: everest-framework 0.19.2 feature/performance-improvements@v0.19.1-35-g64e9ae2
manager          :: Config loading completed in 188ms
manager          :: Starting 16 modules
api:API          :: Module api initialized [53ms]
auth:Auth        :: Module auth initialized [55ms]
connector_1:Evs  :: Module connector_1 initialized [81ms]
connector_1_pow  :: Module connector_1_powerpath initialized [57ms]
energy_manager:  :: Module energy_manager initialized [44ms]
error_history:E  :: Module error_history initialized [61ms]
ev_manager:EvMa  :: Module ev_manager initialized [47ms]
evse_security:E  :: Module evse_security initialized [64ms]
grid_connection  :: Module grid_connection_point initialized [68ms]
iso15118_car     :: Module iso15118_car initialized [45ms]
iso15118_charge  :: Module iso15118_charger initialized [61ms]
persistent_stor  :: Module persistent_store initialized [45ms]
setup:Setup      :: Module setup initialized [38ms]
slac:SlacSimula  :: Module slac initialized [42ms]
token_provider:  :: Module token_provider initialized [38ms]
token_validator  :: Module token_validator initialized [35ms]
manager          :: Clearing retained topics published by manager during startup
manager          :: 🚙🚙🚙 All modules are initialized. EVerest up and running [1700ms] 🚙🚙🚙
```
vs.
```
manager          :: everest-framework 0.19.1 @v0.19.1
manager          :: Config loading completed in 336ms
api:API          :: Module api initialized [84ms]
auth:Auth        :: Module auth initialized [65ms]
connector_1:Evs  :: Module connector_1 initialized [86ms]
energy_manager:  :: Module energy_manager initialized [76ms]
error_history:E  :: Module error_history initialized [94ms]
ev_manager:EvMa  :: Module ev_manager initialized [89ms]
evse_security:E  :: Module evse_security initialized [96ms]
grid_connection  :: Module grid_connection_point initialized [79ms]
iso15118_charge  :: Module iso15118_charger initialized [87ms]
persistent_stor  :: Module persistent_store initialized [63ms]
setup:Setup      :: Module setup initialized [60ms]
slac:SlacSimula  :: Module slac initialized [70ms]
token_provider:  :: Module token_provider initialized [67ms]
token_validator  :: Module token_validator initialized [65ms]
manager          :: 🚙🚙🚙 All modules are initialized. EVerest up and running [1721ms] 🚙🚙🚙
```

Comparing the amount of threads a module uses shows a measurable decrease (refactoring the error handling to use less topics/handlers will have an even big impact on eg. ErrorHistory https://github.com/EVerest/everest-framework/pull/225 )

(v0.19.2 vs v0.19.1)
```
manager            35
controller          2
api:API            22
auth:Auth          22
connector_1:Evs   121
connector_1_pow    15
energy_manager:     6
error_history:E   119
ev_manager:EvMa    16
evse_security:E     6
grid_connection     7
python3             5
iso15118_charge     8
persistent_stor     5
setup:Setup        12
slac:SlacSimula     7
token_provider:    13
token_validator     5
```
vs
```
manager            35
controller          2
api:API            55
auth:Auth          55
connector_1:Evs   156
connector_1_pow    47
energy_manager:    39
error_history:E   151
ev_manager:EvMa    50
evse_security:E    38
grid_connection    40
python3            37
iso15118_charge    41
persistent_stor    37
setup:Setup        45
slac:SlacSimula    39
token_provider:    45
token_validator    37
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,8 @@
 Checks: >
     bugprone*,
     misc-const-correctness,
+    performance*,
+    -performance-avoid-endl,
     -llvmlibc*,
     -fuchsia-overloaded-operator,
     -fuchsia-statically-constructed-objects,
--- a/everestjs/conversions.cpp
+++ b/everestjs/conversions.cpp
@@ -54,6 +54,14 @@ Everest::json convertToJson(const Napi::Value& value) {
                         napi_valuetype_strings[value.Type()]));
 }
 
+Everest::json convertToConfigMap(const Everest::json& json_config) {
+    json config_map;
+    for (auto& entry : json_config.items()) {
+        config_map[entry.key()] = entry.value().at("value");
+    }
+    return config_map;
+}
+
 Everest::TelemetryMap convertToTelemetryMap(const Napi::Object& obj) {
     BOOST_LOG_FUNCTION();
     Everest::TelemetryMap telemetry;
--- a/everestpy/src/everest/everestpy.cpp
+++ b/everestpy/src/everest/everestpy.cpp
@@ -22,10 +22,20 @@ namespace py = pybind11;
 
 PYBIND11_MODULE(everestpy, m) {
 
+    py::class_<Everest::MQTTSettings>(m, "MQTTSettings")
+        .def(py::init<>())
+        .def_readwrite("broker_socket_path", &Everest::MQTTSettings::broker_socket_path)
+        .def_readwrite("broker_host", &Everest::MQTTSettings::broker_host)
+        .def_readwrite("broker_port", &Everest::MQTTSettings::broker_port)
+        .def_readwrite("everest_prefix", &Everest::MQTTSettings::everest_prefix)
+        .def_readwrite("external_prefix", &Everest::MQTTSettings::external_prefix)
+        .def("uses_socket", &Everest::MQTTSettings::uses_socket);
+
     // FIXME (aw): add m.doc?
     py::class_<RuntimeSession>(m, "RuntimeSession")
         .def(py::init<>())
-        .def(py::init<const std::string&, const std::string&>());
+        .def(py::init<const std::string&, const std::string&>())
+        .def(py::init<const Everest::MQTTSettings&, const std::string&>());
 
     py::class_<ModuleInfo::Paths>(m, "ModuleInfoPaths")
         .def_readonly("etc", &ModuleInfo::Paths::etc)
--- a/everestpy/src/everest/everestpy.cpp
+++ b/everestpy/src/everest/everestpy.cpp
@@ -130,7 +130,7 @@ PYBIND11_MODULE(everestpy, m) {
         .def(py::init<const std::string&, const RuntimeSession&>())
         .def("say_hello", &Module::say_hello)
         .def("init_done", py::overload_cast<>(&Module::init_done))
-        .def("init_done", py::overload_cast<std::function<void()>>(&Module::init_done))
+        .def("init_done", py::overload_cast<const std::function<void()>&>(&Module::init_done))
         .def("call_command", &Module::call_command)
         .def("publish_variable", &Module::publish_variable)
         .def("implement_command", &Module::implement_command)
--- a/everestpy/src/everest/misc.cpp
+++ b/everestpy/src/everest/misc.cpp
@@ -82,8 +82,8 @@ RuntimeSession::RuntimeSession() {
 ModuleSetup create_setup_from_config(const std::string& module_id, Everest::Config& config) {
     ModuleSetup setup;
 
-    const std::string& module_name = config.get_main_config().at(module_id).at("module");
-    const auto module_manifest = config.get_manifests().at(module_name);
+    const std::string& module_name = config.get_module_name(module_id);
+    const auto& module_manifest = config.get_manifests().at(module_name);
 
     // setup connections
     for (const auto& requirement : module_manifest.at("requires").items()) {
--- a/everestpy/src/everest/misc.cpp
+++ b/everestpy/src/everest/misc.cpp
@@ -63,18 +63,14 @@ RuntimeSession::RuntimeSession(const std::string& prefix, const std::string& con
     // We extract the settings from the config file so everest-testing doesn't break
     const auto ms = Everest::ManagerSettings(prefix, config_file);
 
-    Everest::Logging::init(ms.runtime_settings->logging_config_file.string());
+    this->logging_config_file = ms.runtime_settings->logging_config_file;
 
     this->mqtt_settings = ms.mqtt_settings;
 }
 
 RuntimeSession::RuntimeSession() {
-    const auto module_id = get_variable_from_env("EV_MODULE");
-
-    namespace fs = std::filesystem;
-    const fs::path logging_config_file =
+    this->logging_config_file =
         Everest::assert_file(get_variable_from_env("EV_LOG_CONF_FILE"), "Default logging config");
-    Everest::Logging::init(logging_config_file.string(), module_id);
 
     this->mqtt_settings = get_mqtt_settings_from_env();
 }
--- a/everestpy/src/everest/misc.cpp
+++ b/everestpy/src/everest/misc.cpp
@@ -107,7 +107,7 @@ ModuleSetup create_setup_from_config(const std::string& module_id, Everest::Conf
             const auto& req_route = req_route_list[i];
             const auto fulfillment =
                 Fulfillment{req_route["module_id"], req_route["implementation_id"], {requirement_id, i}};
-            fulfillment_list.emplace_back(std::move(fulfillment));
+            fulfillment_list.emplace_back(fulfillment);
         }
     }
 
--- a/everestpy/src/everest/misc.cpp
+++ b/everestpy/src/everest/misc.cpp
@@ -53,6 +53,17 @@ static Everest::MQTTSettings get_mqtt_settings_from_env() {
     }
 }
 
+RuntimeSession::RuntimeSession(const Everest::MQTTSettings& mqtt_settings, const std::string& logging_config) {
+    this->mqtt_settings = mqtt_settings;
+    if (logging_config.empty()) {
+        this->logging_config_file = Everest::assert_dir(Everest::defaults::PREFIX, "Default prefix") /
+                                    std::filesystem::path(Everest::defaults::SYSCONF_DIR) /
+                                    Everest::defaults::NAMESPACE / Everest::defaults::LOGGING_CONFIG_NAME;
+    } else {
+        this->logging_config_file = Everest::assert_file(logging_config, "Default logging config");
+    }
+}
+
 /// This is just kept for compatibility
 RuntimeSession::RuntimeSession(const std::string& prefix, const std::string& config_file) {
     EVLOG_warning
--- a/everestpy/src/everest/module.cpp
+++ b/everestpy/src/everest/module.cpp
@@ -23,6 +23,8 @@ Module::Module(const RuntimeSession& session) : Module(get_variable_from_env("EV
 Module::Module(const std::string& module_id_, const RuntimeSession& session_) :
     module_id(module_id_), session(session_), start_time(std::chrono::system_clock::now()) {
 
+    Everest::Logging::init(session.get_logging_config_file().string(), module_id);
+
     this->mqtt_abstraction = std::make_shared<Everest::MQTTAbstraction>(session.get_mqtt_settings());
     this->mqtt_abstraction->connect();
     this->mqtt_abstraction->spawn_main_loop_thread();
--- a/everestpy/src/everest/module.cpp
+++ b/everestpy/src/everest/module.cpp
@@ -12,7 +12,7 @@
 std::unique_ptr<Everest::Everest>
 Module::create_everest_instance(const std::string& module_id, const Everest::Config& config,
                                 const Everest::RuntimeSettings& rs,
-                                std::shared_ptr<Everest::MQTTAbstraction> mqtt_abstraction) {
+                                const std::shared_ptr<Everest::MQTTAbstraction>& mqtt_abstraction) {
     return std::make_unique<Everest::Everest>(module_id, config, rs.validate_schema, mqtt_abstraction,
                                               rs.telemetry_prefix, rs.telemetry_enabled);
 }
--- a/everestpy/src/everest/module.cpp
+++ b/everestpy/src/everest/module.cpp
@@ -21,7 +21,7 @@ Module::Module(const RuntimeSession& session) : Module(get_variable_from_env("EV
 }
 
 Module::Module(const std::string& module_id_, const RuntimeSession& session_) :
-    module_id(module_id_), session(session_) {
+    module_id(module_id_), session(session_), start_time(std::chrono::system_clock::now()) {
 
     this->mqtt_abstraction = std::make_shared<Everest::MQTTAbstraction>(session.get_mqtt_settings());
     this->mqtt_abstraction->connect();
--- a/include/framework/everest.hpp
+++ b/include/framework/everest.hpp
@@ -49,7 +49,7 @@ struct ErrorFactory;
 class Everest {
 public:
     Everest(std::string module_id, const Config& config, bool validate_data_with_schema,
-            std::shared_ptr<MQTTAbstraction> mqtt_abstraction, const std::string& telemetry_prefix,
+            const std::shared_ptr<MQTTAbstraction>& mqtt_abstraction, const std::string& telemetry_prefix,
             bool telemetry_enabled);
 
     // forbid copy assignment and copy construction
@@ -66,7 +66,7 @@ public:
     ///
     /// \brief Allows a module to indicate that it provides the given command \p cmd
     ///
-    void provide_cmd(const std::string impl_id, const std::string cmd_name, const JsonCommand handler);
+    void provide_cmd(const std::string& impl_id, const std::string cmd_name, const JsonCommand& handler);
     void provide_cmd(const cmd& cmd);
 
     ///
@@ -217,7 +217,7 @@ private:
     bool telemetry_enabled;
     std::optional<ModuleTierMappings> module_tier_mappings;
 
-    void handle_ready(nlohmann::json data);
+    void handle_ready(const nlohmann::json& data);
 
     void heartbeat();
 
--- a/include/framework/runtime.hpp
+++ b/include/framework/runtime.hpp
@@ -192,7 +192,8 @@ private:
 
 public:
     explicit ModuleLoader(int argc, char* argv[], ModuleCallbacks callbacks) :
-        ModuleLoader(argc, argv, callbacks, {"undefined project", "undefined version", "undefined git version"}){};
+        ModuleLoader(argc, argv, std::move(callbacks),
+                     {"undefined project", "undefined version", "undefined git version"}){};
     explicit ModuleLoader(int argc, char* argv[], ModuleCallbacks callbacks,
                           const VersionInformation& version_information);
 
--- a/include/framework/runtime.hpp
+++ b/include/framework/runtime.hpp
@@ -194,7 +194,7 @@ public:
     explicit ModuleLoader(int argc, char* argv[], ModuleCallbacks callbacks) :
         ModuleLoader(argc, argv, callbacks, {"undefined project", "undefined version", "undefined git version"}){};
     explicit ModuleLoader(int argc, char* argv[], ModuleCallbacks callbacks,
-                          const VersionInformation version_information);
+                          const VersionInformation& version_information);
 
     int initialize();
 };
--- a/include/utils/config.hpp
+++ b/include/utils/config.hpp
@@ -30,12 +30,25 @@ struct RuntimeSettings;
 ///
 /// \brief A structure that contains all available schemas
 ///
-struct schemas {
+struct Schemas {
     nlohmann::json config;                 ///< The config schema
     nlohmann::json manifest;               ///< The manifest scheme
     nlohmann::json interface;              ///< The interface schema
     nlohmann::json type;                   ///< The type schema
     nlohmann::json error_declaration_list; ///< The error-declaration-list schema
+};
+
+struct Validators {
+    std::unique_ptr<nlohmann::json_schema::json_validator> config;
+    std::unique_ptr<nlohmann::json_schema::json_validator> manifest;
+    std::unique_ptr<nlohmann::json_schema::json_validator> type;
+    std::unique_ptr<nlohmann::json_schema::json_validator> interface;
+    std::unique_ptr<nlohmann::json_schema::json_validator> error_declaration_list;
+};
+
+struct SchemaValidation {
+    Schemas schemas;
+    Validators validators;
 };
 
 ///
@@ -51,6 +64,17 @@ struct ImplementationInfo {
 };
 
 ///
+/// \brief A simple json schema loader that uses the builtin draft7 schema of
+/// the json schema validator when it encounters it, throws an exception
+/// otherwise
+void loader(const nlohmann::json_uri& uri, nlohmann::json& schema);
+
+///
+/// \brief An extension to the default format checker of the json schema
+/// validator supporting uris
+void format_checker(const std::string& format, const std::string& value);
+
+///
 /// \brief Base class for configs
 ///
 class ConfigBase {
@@ -62,7 +86,7 @@ protected:
     nlohmann::json interfaces;
     nlohmann::json interface_definitions;
     nlohmann::json types;
-    schemas _schemas;
+    Schemas schemas;
 
     std::unordered_map<std::string, ModuleTierMappings> tier_mappings;
     // experimental caches
@@ -190,6 +214,8 @@ class ManagerConfig : public ConfigBase {
 private:
     const ManagerSettings& ms;
     std::unordered_map<std::string, std::optional<TelemetryConfig>> telemetry_configs;
+    Validators validators;
+    std::unique_ptr<nlohmann::json_schema::json_validator> draft7_validator;
 
     ///
     /// \brief loads and validates the manifest of the module \p module_id using the provided \p module config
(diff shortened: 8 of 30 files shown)


everestpy/src/everest/everestpy.cpp
```diff
@@ -130,7 +130,7 @@ PYBIND11_MODULE(everestpy, m) {
         .def(py::init<const std::string&, const RuntimeSession&>())
         .def("say_hello", &Module::say_hello)
         .def("init_done", py::overload_cast<>(&Module::init_done))
-        .def("init_done", py::overload_cast<std::function<void()>>(&Module::init_done))
+        .def("init_done", py::overload_cast<const std::function<void()>&>(&Module::init_done))
```
a-w50, line 143: This might be a matter of taste.  I'm just wondering if now any appearance of any `std::function` will be passed by reference instead of value?

everestpy/src/everest/module.cpp
```diff
@@ -12,7 +12,7 @@
 std::unique_ptr<Everest::Everest>
 Module::create_everest_instance(const std::string& module_id, const Everest::Config& config,
                                 const Everest::RuntimeSettings& rs,
-                                std::shared_ptr<Everest::MQTTAbstraction> mqtt_abstraction) {
+                                const std::shared_ptr<Everest::MQTTAbstraction>& mqtt_abstraction) {
```
a-w50, line 15: A constant reference to a shared pointer is almost the same as the pointer itself.  `Everest::MQTTAbstraction&` would be the better choice here.

include/framework/everest.hpp
```diff
@@ -66,7 +66,7 @@ class Everest {
     ///
     /// \brief Allows a module to indicate that it provides the given command \p cmd
     ///
-    void provide_cmd(const std::string impl_id, const std::string cmd_name, const JsonCommand handler);
+    void provide_cmd(const std::string& impl_id, const std::string cmd_name, const JsonCommand& handler);
```
a-w50, line 69: Why const ref'ing only `impl_id` and not `cmd_name`?

include/framework/runtime.hpp
```diff
@@ -194,7 +194,7 @@ class ModuleLoader {
     explicit ModuleLoader(int argc, char* argv[], ModuleCallbacks callbacks) :
         ModuleLoader(argc, argv, callbacks, {"undefined project", "undefined version", "undefined git version"}){};
     explicit ModuleLoader(int argc, char* argv[], ModuleCallbacks callbacks,
-                          const VersionInformation version_information);
+                          const VersionInformation& version_information);
```
a-w50, line 198: This should potentially passed by value, as this information is probably only moved to the `ModuleLoader`.

include/utils/message_queue.hpp
```diff
@@ -79,10 +79,10 @@ class MessageHandler {
     void stop();
 
     /// \brief Adds a \p handler that will receive messages from the queue.
-    void add_handler(std::shared_ptr<TypedHandler> handler);
+    void add_handler(const std::shared_ptr<TypedHandler>& handler);
```
a-w50, line 82: See my comment above - I think this might not be a good choice.

lib/config.cpp
```diff
@@ -78,8 +78,8 @@ static ParsedConfigMap parse_config_map(const json& config_map_schema, const jso
 
     // validate each config entry
     for (const auto& config_entry_el : config_map_schema.items()) {
-        const std::string config_entry_name = config_entry_el.key();
-        const json config_entry = config_entry_el.value();
+        const std::string& config_entry_name = config_entry_el.key();
+        const json& config_entry = config_entry_el.value();
```
a-w50, line 160: `const auto&`

lib/config.cpp
```diff
@@ -579,7 +579,7 @@ void ManagerConfig::load_and_validate_manifest(const std::string& module_id, con
 
     // validate config for !module
     {
-        const json config_map = module_config.at("config_module");
+        const json& config_map = module_config.at("config_module");
```
a-w50, line 660: `const auto&`

lib/config.cpp
```diff
@@ -1167,7 +1167,7 @@ ModuleConfigs Config::get_module_configs(const std::string& module_id) const {
             for (const auto& entry : conf_map.value().items()) {
                 const json entry_type = config_schema.at(entry.key()).at("type");
                 ConfigEntry value;
-                const json data = entry.value();
+                const json& data = entry.value();
```
a-w50: `const auto&`

lib/everest.cpp
```diff
@@ -71,7 +71,8 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
     // setup error_manager_req_global if enabled + error_database + error_state_monitor
     if (this->module_manifest.contains("enable_global_errors") &&
         this->module_manifest.at("enable_global_errors").get<bool>()) {
-        std::shared_ptr<error::ErrorDatabaseMap> global_error_database = std::make_shared<error::ErrorDatabaseMap>();
+        const std::shared_ptr<error::ErrorDatabaseMap>& global_error_database =
```
a-w50, line 74: `const auto&`

lib/everest.cpp
```diff
@@ -90,7 +91,7 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
     // setup error_managers, error_state_monitors, error_factories and error_databases for all implementations
     for (const std::string& impl : Config::keys(this->module_manifest.at("provides"))) {
         // setup shared database
-        std::shared_ptr<error::ErrorDatabaseMap> error_database = std::make_shared<error::ErrorDatabaseMap>();
+        const std::shared_ptr<error::ErrorDatabaseMap> error_database = std::make_shared<error::ErrorDatabaseMap>();
```
a-w50, line 94: `const auto&`

lib/everest.cpp
```diff
@@ -666,13 +667,13 @@ void Everest::subscribe_global_all_errors(const error::ErrorCallback& callback,
...
-            const std::string impl_id = impl.key();
+            const std::string& impl_id = impl.key();
             const std::string interface = impl.value().at("interface");
             const json errors = this->config.get_interface_definition(interface).at("errors");
             for (const auto& error_namespace_it : errors.items()) {
-                const std::string error_type_namespace = error_namespace_it.key();
+                const std::string& error_type_namespace = error_namespace_it.key();
                 for (const auto& error_name_it : error_namespace_it.value().items()) {
-                    const std::string error_type_name = error_name_it.key();
+                    const std::string& error_type_name = error_name_it.key();
```
a-w50, line 675: There are 4 nested for loops here, I would appreciate, that if places like these get touched, that they get refactored or a FIXME gets added.

lib/message_queue.cpp
```diff
@@ -70,7 +70,7 @@ MessageHandler::MessageHandler() : running(true) {
             std::vector<std::shared_ptr<TypedHandler>> local_handlers;
             {
                 const std::lock_guard<std::mutex> handlers_lock(handler_list_mutex);
-                for (auto handler : this->handlers) {
+                for (const auto& handler : this->handlers) {
                     local_handlers.push_back(handler);
                 }
```
a-w50, line 75: Wouldn't be `local_handlers = this->handlers;` be enough?

.clang-tidy
a-w50, line 1: Is there any documentation on how to run clang-tidy on everest-framework?

src/manager.cpp
```diff
@@ -352,12 +352,13 @@ static std::map<pid_t, std::string> start_modules(ManagerConfig& config, MQTTAbs
     mqtt_abstraction.publish(fmt::format("{}module_config_cache", ms.mqtt_settings.everest_prefix), module_config_cache,
                              QOS::QOS2, true);
 
-    for (const auto& module : serialized_config.at("module_names").items()) {
-        const std::string module_name = module.key();
-        json serialized_mod_config = serialized_config;
+    for (const auto& module_name_entry : module_names) {
+        const auto& module_name = module_name_entry.first;
+        const auto& module_type = module_name_entry.second;
```
a-w50, line 346: Structured binding might be possible here.

src/manager.cpp
```diff
@@ -352,12 +352,13 @@ static std::map<pid_t, std::string> start_modules(ManagerConfig& config, MQTTAbs
...
+        json serialized_mod_config = json::object({{"module_names", module_names}});
         serialized_mod_config["module_config"] = json::object();
+        serialized_mod_config["module_config"][module_name] = main_config.at(module_name);
         // add mappings of fulfillments
-        serialized_mod_config["module_config"][module_name] = serialized_config.at("main").at(module_name);
```
a-w50: This refactoring conceptually improves the code - it might become even more readable and testable, if the module config generation would get its own function

lib/config.cpp
```diff
@@ -1249,21 +1274,33 @@ void Config::ref_loader(const json_uri& uri, json& schema) {
...
+    auto [type_schema, type_val] = Config::load_schema(schemas_dir / "type.yaml");
+    schema_validation.schemas.type = type_schema;
+    schema_validation.validators.type = std::move(type_val);
+    auto [error_declaration_list_schema, error_declaration_list_val] =
+        Config::load_schema(schemas_dir / "error-declaration-list.yaml");
+    schema_validation.schemas.error_declaration_list = error_declaration_list_schema;
+    schema_validation.validators.error_declaration_list = std::move(error_declaration_list_val);
+
+    return schema_validation;
 }
```
a-w50: Couldn't this be a free function?

lib/config.cpp
```diff
@@ -339,7 +367,7 @@ const json& ConfigBase::get_settings() const {
 
 const json ConfigBase::get_schemas() const {
```
a-w50, line 420: This should probably return a reference.

include/utils/config.hpp
```diff
@@ -30,14 +30,27 @@ struct RuntimeSettings;
...
     nlohmann::json type;                   ///< The type schema
     nlohmann::json error_declaration_list; ///< The error-declaration-list schema
 };
 
+struct Validators {
+    std::unique_ptr<nlohmann::json_schema::json_validator> config;
+    std::unique_ptr<nlohmann::json_schema::json_validator> manifest;
+    std::unique_ptr<nlohmann::json_schema::json_validator> type;
+    std::unique_ptr<nlohmann::json_schema::json_validator> interface;
+    std::unique_ptr<nlohmann::json_schema::json_validator> error_declaration_list;
```
a-w50: Looking a the class `json_schema::json_validator`, it already only contains a `unique_ptr`.  So you don't need to put it into a `unique_ptr` again.  Just use it as a regular value type and `std::move` it.

lib/yaml_loader.cpp
```diff
@@ -78,6 +78,16 @@ static nlohmann::ordered_json ryml_to_nlohmann_json(const c4::yml::NodeRef& ryml
...
 }
 
+static std::string load_file_content(const std::filesystem::path& path) {
+    std::ifstream ifs(path.string());
+    ifs.seekg(0, std::ios::end);
+    std::string content;
+    content.reserve(ifs.tellg());
+    ifs.seekg(0);
+    content.assign(std::istreambuf_iterator<char>(ifs), std::istreambuf_iterator<char>());
+    return content;
```
a-w50: This is probably too much optimization.  The previous version was easier to read - but having a separate function is completely valid.

lib/yaml_loader.cpp
```diff
@@ -4,20 +4,26 @@
...
 #include <ryml.hpp>
 #include <ryml_std.hpp>
 
 #include <everest/logging.hpp>
 
+static std::streamsize clamp(std::size_t len) {
+    return (len <= std::numeric_limits<std::streamsize>::max()) ? static_cast<std::streamsize>(len)
+                                                                : std::numeric_limits<std::streamsize>::max();
+}
+
```
a-w50, line 19: This function could potentially made `constexpr`.  Nevertheless, removing the narrowing warning by clamping the value seems a bit strange.  But somehow both interfaces (stringstream and yaml error info) need to be glued together.  Still I would prefer to have a templated function (using `T` for `std::streamsize`), and renaming it to `clamp_to`, `restrict_to` or `restrict_length`, as it reads a bit better: `clamp_to<std::streamsize>(len)`.

include/utils/mqtt_abstraction_impl.hpp
```diff
@@ -20,7 +20,7 @@
 
 #include <utils/thread.hpp>
 
-constexpr std::size_t MQTT_BUF_SIZE = 500 * 1024;
+constexpr std::size_t MQTT_BUF_SIZE = std::size_t{500} * std::size_t{1024};
```
a-w50, line 23: how about `constexpr auto MQTT_BUF_SIZE = std::size_t{500 * 1024}`?

lib/module_config.cpp
```diff
@@ -104,6 +104,10 @@ json get_module_config(const std::shared_ptr<MQTTAbstraction>& mqtt, const std::
     const auto module_config_cache = mqtt->get(module_config_cache_topic, QOS::QOS2);
     result["module_config_cache"] = module_config_cache;
 
+    const auto module_names_topic = fmt::format("{}module_names", everest_prefix);
```
a-w50: Should be a constant somewhere, not hard coded here.

include/utils/config.hpp
```diff
@@ -313,7 +313,7 @@ class Config : public ConfigBase {
 
     ///
     /// \returns the commands that the modules \p module_name implements from the given implementation \p impl_id
-    nlohmann::json get_module_cmds(const std::string& module_name, const std::string& impl_id);
+    const nlohmann::json& get_module_cmds(const std::string& module_name, const std::string& impl_id);
```
a-w50, line 329: These functions should be `const` too

lib/config.cpp
```diff
@@ -1181,14 +1181,12 @@ ModuleConfigs Config::get_module_configs(const std::string& module_id) const {
...
-                                           ? manifest.at("config")
-                                           : manifest.at("provides").at(conf_map.key()).at("config");
             ConfigMap processed_conf_map;
             for (const auto& entry : conf_map.value().items()) {
-                const json entry_type = config_schema.at(entry.key()).at("type");
+                const auto& entry_value = entry.value();
+                const json entry_type = entry_value.at("type");
                 ConfigEntry value;
-                const json& data = entry.value();
+                const json& data = entry_value.at("value");
```
a-w50, line 1240: This isn't the same as before, anymore, right?  Is this intended?

lib/config.cpp
```diff
@@ -1181,14 +1181,12 @@ ModuleConfigs Config::get_module_configs(const std::string& module_id) const {
...
 
         for (const auto& conf_map : config_maps.items()) {
-            const json config_schema = (conf_map.key() == "!module")
-                                           ? manifest.at("config")
-                                           : manifest.at("provides").at(conf_map.key()).at("config");
             ConfigMap processed_conf_map;
             for (const auto& entry : conf_map.value().items()) {
-                const json entry_type = config_schema.at(entry.key()).at("type");
+                const auto& entry_value = entry.value();
+                const json entry_type = entry_value.at("type");
```
a-w50: Why no `auto&` here?

lib/module_config.cpp
```diff
@@ -92,8 +92,18 @@ json get_module_config(const std::shared_ptr<MQTTAbstraction>& mqtt, const std::
...
+
     const auto manifests_topic = fmt::format("{}manifests", everest_prefix);
-    const auto manifests = mqtt->get(manifests_topic, QOS::QOS2);
+    auto manifests = json::object();
+    for (const auto& module_name : module_names) {
+        auto manifest_topic = fmt::format("{}manifests/{}", everest_prefix, module_name.get<std::string>());
+        auto manifest = mqtt->get(manifest_topic, QOS::QOS2);
+        manifests[module_name] = manifest;
+    }
+
```
a-w50: These hardcoded values like `module_names` and `manifest` should be defined in some central place.

lib/module_config.cpp
```diff
@@ -92,8 +92,18 @@ json get_module_config(const std::shared_ptr<MQTTAbstraction>& mqtt, const std::
...
+    const auto module_names_topic = fmt::format("{}module_names", everest_prefix);
+    const auto module_names = mqtt->get(module_names_topic, QOS::QOS2);
+    result["module_names"] = module_names;
+
     const auto manifests_topic = fmt::format("{}manifests", everest_prefix);
-    const auto manifests = mqtt->get(manifests_topic, QOS::QOS2);
+    auto manifests = json::object();
+    for (const auto& module_name : module_names) {
+        auto manifest_topic = fmt::format("{}manifests/{}", everest_prefix, module_name.get<std::string>());
+        auto manifest = mqtt->get(manifest_topic, QOS::QOS2);
```
a-w50: This is a no-go in my point of view. I didn't see, that it is has been merged like this before.
But getting the module config shouldn't dynamically read values via mqtt.  This should be done during construction for several reasons.

include/utils/mqtt_abstraction.hpp
```diff
@@ -64,6 +64,10 @@ class MQTTAbstraction {
     /// \copydoc MQTTAbstractionImpl::unsubscribe(const std::string&)
     void unsubscribe(const std::string& topic);
 
+    ///
+    /// \copydoc MQTTAbstractionImpl::get_async(const std::string&, QOS)
+    AsyncReturn get_async(const std::string& topic, QOS qos);
```
a-w50: This `get_async` implementation needs to be discussed separately.

everestjs/conversions.cpp
```diff
@@ -54,6 +54,14 @@ Everest::json convertToJson(const Napi::Value& value) {
...
 }
 
+Everest::json convertToConfigMap(const Everest::json& json_config) {
+    json config_map;
+    for (auto& entry : json_config.items()) {
+        config_map[entry.key()] = entry.value().at("value");
+    }
+    return config_map;
+}
+
```
a-w50, line 64: This looks like a data layout problem.  Not sure, if this additional data restructuring is really necessary.

lib/module_config.cpp
```diff
@@ -15,40 +16,71 @@ namespace Everest {
...
+    json result;
+    if (future_status == std::future_status::timeout) {
+        mqtt->unregister_handler(topic, token);
+        EVLOG_AND_THROW(EverestTimeoutError(fmt::format("Timeout while waiting for result of get({})", topic)));
+    }
+    if (future_status == std::future_status::ready) {
+        return future.get();
+    }
+
+    return result;
```
a-w50: needs discussion

src/manager.cpp
```diff
@@ -638,7 +638,7 @@ int boot(const po::variables_map& vm) {
     EVLOG_info << ms.version_information;
     EVLOG_info << "";
 
-    if (ms.mqtt_settings.uses_socket()) {
+    if (not ms.mqtt_settings.uses_socket()) {
```
a-w50: This is a bugfix, not a performance improvement.  I would like to have this as a separate commit.

src/manager.cpp
```diff
@@ -400,6 +400,17 @@ static std::map<pid_t, std::string> start_modules(ManagerConfig& config, MQTTAbs
...
+        const Handler module_get_config_handler = [module_name, config_topic, serialized_mod_config,
+                                                   &mqtt_abstraction](const std::string&, const nlohmann::json& json) {
+            mqtt_abstraction.publish(config_topic, serialized_mod_config.dump(), QOS::QOS2);
+        };
+
+        const std::string get_config_topic = fmt::format("{}/get_config", config.mqtt_module_prefix(module_name));
+        module_it->second.get_config_token = std::make_shared<TypedHandler>(
+            HandlerType::ExternalMQTT, std::make_shared<Handler>(module_get_config_handler));
+        mqtt_abstraction.register_handler(get_config_topic, module_it->second.get_config_token, QOS::QOS2);
+
```
a-w50, line 389: What is the reason for this rearrangement?

everestpy/src/everest/misc.cpp
```diff
@@ -63,18 +63,14 @@ RuntimeSession::RuntimeSession(const std::string& prefix, const std::string& con
     // We extract the settings from the config file so everest-testing doesn't break
     const auto ms = Everest::ManagerSettings(prefix, config_file);
 
-    Everest::Logging::init(ms.runtime_settings->logging_config_file.string());
```
a-w50, line 66: Why has this been moved?  Isn't logging already used before until the module is constructed?

everestpy/src/everest/everestpy.cpp
```diff
@@ -22,10 +22,20 @@ namespace py = pybind11;
...
 
+    py::class_<Everest::MQTTSettings>(m, "MQTTSettings")
+        .def(py::init<>())
+        .def_readwrite("broker_socket_path", &Everest::MQTTSettings::broker_socket_path)
+        .def_readwrite("broker_host", &Everest::MQTTSettings::broker_host)
+        .def_readwrite("broker_port", &Everest::MQTTSettings::broker_port)
+        .def_readwrite("everest_prefix", &Everest::MQTTSettings::everest_prefix)
+        .def_readwrite("external_prefix", &Everest::MQTTSettings::external_prefix)
+        .def("uses_socket", &Everest::MQTTSettings::uses_socket);
+
```
a-w50, line 33: Why has this been added?  The more we broaden the api, the harder it gets to refactor.

include/utils/mqtt_abstraction_impl.hpp
```diff
@@ -80,6 +80,10 @@ class MQTTAbstractionImpl {
     /// \brief unsubscribes from the given \p topic
     void unsubscribe(const std::string& topic);
 
+    ///
+    /// \brief clears any previously published topics that had the retain flag set
+    void clear_retained_topics();
```
a-w50, line 85: Is there any documentation on how or when the `retain-topics` flag should be used?

src/manager.cpp
```diff
@@ -749,6 +753,9 @@ int boot(const po::variables_map& vm) {
 
     auto module_handles =
         start_modules(*config, mqtt_abstraction, ignored_modules, standalone_modules, ms, status_fifo, retain_topics);
+    if (module_handles.empty()) {
+        return EXIT_FAILURE;
```
a-w50, line 757: Should this be considered a failure?

a-w50 (commented):
These are quite a lot of commits. I would appreciate, if these could be split into several PRs, focused on their real use case like refactor, bugfix, features etc...

lib/everest.cpp
hikinggrass: This gets refactored in https://github.com/EVerest/everest-framework/pull/225

include/framework/everest.hpp
hikinggrass: Addressed in https://github.com/EVerest/everest-framework/pull/227

include/framework/runtime.hpp
hikinggrass: Addressed in https://github.com/EVerest/everest-framework/pull/227

include/utils/mqtt_abstraction_impl.hpp
hikinggrass: `std::size_t{500 * 1024}` still produces a widening conversion warning since 500 and 1024 are interpreted as string. In C++23 we could postfix it with `uz` but I think this way is fine for now (it only fixes a warning anyway). Addressed in https://github.com/EVerest/everest-framework/pull/227

lib/yaml_loader.cpp
hikinggrass: Addressed in https://github.com/EVerest/everest-framework/pull/227

lib/yaml_loader.cpp
hikinggrass: Agree, I did some quick comparisons between the two and didn't really see much different one way or the other so I reverted this back to the original code

src/manager.cpp
hikinggrass: Addressed in https://github.com/EVerest/everest-framework/pull/228

lib/config.cpp
hikinggrass: yes, I've turned both `load_schema` and `load_schemas` into free functions

include/utils/config.hpp
hikinggrass: done

include/utils/mqtt_abstraction.hpp
hikinggrass: I moved this (and get() as well) into module_config.cpp since it is only used there

lib/mqtt_abstraction_impl.cpp
```diff
@@ -484,6 +472,7 @@ void MQTTAbstractionImpl::unregister_handler(const std::string& topic, const Tok
             EVLOG_verbose << fmt::format("Unsubscribing from {}", topic);
             this->unsubscribe(topic);
         }
+        this->message_handlers.erase(topic);
```
hikinggrass, line 475: Opened a new PR for this change which helps reduce the amount of idle threads quite significantly: https://github.com/EVerest/everest-framework/pull/229
In general this design needs to be refactored, but that's probably best done in a future PR

include/utils/message_queue.hpp
hikinggrass: I've addressed this (by reverting the change and blocking the suggestion of adding const ref to shared_ptr args) in https://github.com/EVerest/everest-framework/pull/227
I think this was changed by me just because I wasn't paying too close attention to the recommendation of clang-tidy here

everestpy/src/everest/module.cpp
hikinggrass: Here it would have probably been fine because Module will outlive Everest, but I've reverted the change in  https://github.com/EVerest/everest-framework/pull/227

src/manager.cpp
hikinggrass: My thinking was that since the spawned module will first call `get_config` and later on `ready` this should also be reflected in the code here

everestpy/src/everest/everestpy.cpp
hikinggrass: This allows everest-testing to use the MQTT based config distribution mechanism since it's (as far as I know) almost impossible to set the environment variables in the pytest environment. Usage is implemented here: https://github.com/EVerest/everest-utils/pull/173

src/manager.cpp
hikinggrass: That's certainly debatable. The current behavior is that everest will happily load empty configs, but it doesn't signal a completed startup via the status fifo or logs. So for a user it just "hangs" without doing anything else. I thought failing early would be better in this case, but we can also leave the current behavior in place if that's more desirable

PR #225, opened 2025-01-03
Reduce amount of error topics to one per impl

hikinggrass:
Removing the split between error raising and error clearing topics helps with preserving the order of these operations

Reducing the amount of topics, and therefore handlers, also reduces the amount of threads a module using error handling is spawning considerably

### DIFF
--- a/lib/everest.cpp
+++ b/lib/everest.cpp
@@ -513,7 +513,7 @@ void Everest::subscribe_var(const Requirement& req, const std::string& var_name,
 }
 
 void Everest::subscribe_error(const Requirement& req, const error::ErrorType& error_type,
-                              const error::ErrorCallback& callback, const error::ErrorCallback& clear_callback) {
+                              const error::ErrorCallback& raise_callback, const error::ErrorCallback& clear_callback) {
     BOOST_LOG_FUNCTION();
 
     EVLOG_debug << fmt::format("subscribing to error: {}:{}", req.id, error_type);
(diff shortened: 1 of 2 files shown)


lib/everest.cpp
```diff
@@ -549,36 +549,39 @@ void Everest::subscribe_error(const Requirement& req, const error::ErrorType& er
...
 
-    const std::string raise_topic =
-        fmt::format("{}/error/{}", this->config.mqtt_prefix(requirement_module_id, requirement_impl_id), error_type);
+        switch (error.state) {
+        case error::State::Active:
+            EVLOG_debug << fmt::format("Incoming error {}->{}",
+                                       this->config.printable_identifier(requirement_module_id, requirement_impl_id),
+                                       error_type);
+
+            callback(error);
```
andistorm: would be nice to rename this function parameter to raise_callback or so. This  way having `callback` and `clear_callback` in one function is a bit confusing

lib/everest.cpp
```diff
@@ -647,64 +650,49 @@ void Everest::subscribe_global_all_errors(const error::ErrorCallback& callback,
...
+            break;
+        case error::State::ClearedByModule:
+        case error::State::ClearedByReboot:
+            EVLOG_debug << fmt::format(
+                "Incoming error cleared {}->{}",
+                this->config.printable_identifier(error.origin.module_id, error.origin.implementation_id), error.type);
+            clear_callback(error);
+            break;
+        }
```
andistorm: This part of the handler can probably be shared with the other one

andistorm (changes requested):
LGTM in general. Two things I would like to see changed.
I didn't run it but tests should cover this, code looks fine to me.

lib/everest.cpp
hikinggrass: done

andistorm (approved):
🐧

PR #226, opened 2025-01-06
Consistent version number in bindings

hikinggrass:
- everestpy: replace hardcoded version with version number that's ultimately populated by CMake
- everestjs: generate package.json containing the version information with CMake
- everestrs: TODO, since this is not just built by CMake, but also Bazel and Cargo, this should be kept in-sync manually, maybe enforced by a linter rule

### DIFF
--- a/everestjs/.gitignore
+++ b/everestjs/.gitignore
@@ -1,2 +1,3 @@
 *node_modules
+package.json
 package-lock.json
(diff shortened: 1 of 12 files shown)


hikinggrass:
(this is just an experiment, I think we need a better solution)

PR #227, opened 2025-01-07
Various clang tidy recommendations

hikinggrass:
- More const ref usage
- fix narrowing and widening conversion warnings


include/utils/mqtt_abstraction_impl.hpp
```diff
@@ -20,7 +20,7 @@
 
 #include <utils/thread.hpp>
 
-constexpr std::size_t MQTT_BUF_SIZE = 500 * 1024;
+constexpr std::size_t MQTT_BUF_SIZE = std::size_t{500} * std::size_t{1024};
```
a-w50: doesn't
`constexpr auto MQTT_BUF_SIZE = std::size_t(500*1024};` also work?

lib/everest.cpp
```diff
@@ -71,7 +71,8 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
     // setup error_manager_req_global if enabled + error_database + error_state_monitor
     if (this->module_manifest.contains("enable_global_errors") &&
         this->module_manifest.at("enable_global_errors").get<bool>()) {
-        std::shared_ptr<error::ErrorDatabaseMap> global_error_database = std::make_shared<error::ErrorDatabaseMap>();
+        const std::shared_ptr<error::ErrorDatabaseMap>& global_error_database =
```
a-w50: This shouldn't be a reference, also `auto` can be used here.

lib/everest.cpp
```diff
@@ -90,7 +91,7 @@ Everest::Everest(std::string module_id_, const Config& config_, bool validate_da
     // setup error_managers, error_state_monitors, error_factories and error_databases for all implementations
     for (const std::string& impl : Config::keys(this->module_manifest.at("provides"))) {
         // setup shared database
-        std::shared_ptr<error::ErrorDatabaseMap> error_database = std::make_shared<error::ErrorDatabaseMap>();
+        const std::shared_ptr<error::ErrorDatabaseMap> error_database = std::make_shared<error::ErrorDatabaseMap>();
```
a-w50: `auto`

lib/message_queue.cpp
```diff
@@ -70,7 +70,7 @@ MessageHandler::MessageHandler() : running(true) {
             std::vector<std::shared_ptr<TypedHandler>> local_handlers;
             {
                 const std::lock_guard<std::mutex> handlers_lock(handler_list_mutex);
-                for (auto handler : this->handlers) {
+                for (const auto& handler : this->handlers) {
```
a-w50: use `local_handlers = this->handlers`

a-w50 (changes requested):

lib/message_queue.cpp
hikinggrass: this->handlers is a unordered_set, so just a copy-assignment to the local_handlers vector doesn't work, that's probably why this is done this way

include/utils/mqtt_abstraction_impl.hpp
hikinggrass: ah, yeah that works :+1:

lib/everest.cpp
hikinggrass: fixed

lib/everest.cpp
hikinggrass: changed to auto

a-w50 (approved):

PR #229, opened 2025-01-07
Erase MessageHandler for topics without any registered handlers

hikinggrass:
This drastically reduces the amount of lingering threads just hanging around and doing absolutely nothing

### DIFF
--- a/lib/mqtt_abstraction_impl.cpp
+++ b/lib/mqtt_abstraction_impl.cpp
@@ -364,11 +364,10 @@ void MQTTAbstractionImpl::on_mqtt_message(const Message& message) {
         }
         lock.unlock();
 
-        // TODO(kai): this should maybe not even be a warning since it can happen that we unsubscribe from a topic and
-        // have removed the message handler but the MQTT unsubscribe didn't complete yet and we still receive messages
-        // on this topic that we can just ignore
+        // It can happen that we unsubscribe from a topic and have removed the message handler but the MQTT unsubscribe
+        // didn't complete yet and we still receive messages on this topic that we can just ignore
         if (!found) {
-            EVLOG_warning << fmt::format("Internal error: topic '{}' should have a matching handler!", topic);
+            EVLOG_verbose << fmt::format("Topic '{}' should have a matching handler!", topic);
         }
     } catch (boost::exception& e) {
         EVLOG_critical << fmt::format("Caught MQTT on_message boost::exception:\n{}",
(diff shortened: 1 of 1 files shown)


lib/mqtt_abstraction_impl.cpp
```diff
@@ -364,9 +364,11 @@ void MQTTAbstractionImpl::on_mqtt_message(const Message& message) {
         }
         lock.unlock();
 
+        // TODO(kai): this should maybe not even be a warning since it can happen that we unsubscribe from a topic and
+        // have removed the message handler but the MQTT unsubscribe didn't complete yet and we still receive messages
+        // on this topic that we can just ignore
         if (!found) {
-            EVLOG_AND_THROW(
-                EverestInternalError(fmt::format("Internal error: topic '{}' should have a matching handler!", topic)));
+            EVLOG_warning << fmt::format("Internal error: topic '{}' should have a matching handler!", topic);
```
a-w50: This is not really an error, as is written in the comment, this is an valid but unlikely event.

lib/mqtt_abstraction_impl.cpp
```diff
@@ -451,15 +453,15 @@ void MQTTAbstractionImpl::register_handler(const std::string& topic, std::shared
     if (this->message_handlers.count(topic) == 0) {
         this->message_handlers.emplace(std::piecewise_construct, std::forward_as_tuple(topic), std::forward_as_tuple());
     }
-    this->message_handlers[topic].add_handler(handler);
+    this->message_handlers.at(topic).add_handler(handler);
 
     // only subscribe for this topic if we aren't already and the mqtt client is connected
     // if we are not connected the on_mqtt_connect() callback will subscribe to the topic
-    if (this->mqtt_is_connected && this->message_handlers[topic].count_handlers() == 1) {
+    if (this->mqtt_is_connected && this->message_handlers.at(topic).count_handlers() == 1) {
```
a-w50: Seeing this the first time, why should `count_handlers() == 1`?

hikinggrass: I think that's because if count_handler == 1 after registering a new handler it was the first handler being registered which should then trigger a subscribe

lib/mqtt_abstraction_impl.cpp
hikinggrass: Changed to a verbose log and slightly reworded the comment

a-w50 (approved):

PR #231, opened 2025-01-08
Additional debug logging

hikinggrass:
- Log module startup time for JavaScript and Python modules as well 
- Log the number of modules started

### DIFF
--- a/everestpy/src/everest/module.hpp
+++ b/everestpy/src/everest/module.hpp
@@ -3,6 +3,7 @@
 #ifndef EVERESTPY_MODULE_HPP
 #define EVERESTPY_MODULE_HPP
 
+#include <chrono>
 #include <deque>
 #include <functional>
 #include <map>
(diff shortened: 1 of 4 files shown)


everestpy/src/everest/module.hpp
```diff
@@ -28,6 +29,11 @@ class Module {
             handle->register_on_ready_handler(on_ready_handler);
         }
 
+        const auto end_time = std::chrono::system_clock::now();
+        EVLOG_info << "Module " << fmt::format(Everest::TERMINAL_STYLE_BLUE, "{}", this->module_id) << " initialized ["
+                   << std::chrono::duration_cast<std::chrono::milliseconds>(end_time - this->start_time).count()
+                   << "ms]";
+
```
a-w50, line 36: Not sure why this function needs to be defined in the header, should probably move to a cpp file.

a-w50 (approved):

PR #232, opened 2025-01-08
Additional everestpy runtime settings ctor

hikinggrass:
Add an additional RuntimeSettings ctor to everestpy that accepts a MQTTSettings object

This allows eg. everest-testing to directly set the required MQTT connectivity information without having to specify environment variables.

Move init of logging into Module ctor since the module id is only available there

### DIFF
--- a/everestpy/src/everest/module.cpp
+++ b/everestpy/src/everest/module.cpp
@@ -23,6 +23,8 @@ Module::Module(const RuntimeSession& session) : Module(get_variable_from_env("EV
 Module::Module(const std::string& module_id_, const RuntimeSession& session_) :
     module_id(module_id_), session(session_) {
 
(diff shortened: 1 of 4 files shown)


everestpy/src/everest/module.cpp
```diff
@@ -23,6 +23,8 @@ Module::Module(const RuntimeSession& session) : Module(get_variable_from_env("EV
 Module::Module(const std::string& module_id_, const RuntimeSession& session_) :
     module_id(module_id_), session(session_) {
 
+    Everest::Logging::init(session.get_logging_config_file().string(), module_id);
+
```
hikinggrass, line 27: This is debatable, since logging should not happen before this is called. There's one edge-case where logging might happen earlier: in the deprecated RuntimeSession ctor ManagerSettings could log if something goes wrong

PR #233, opened 2025-01-08
Add function to cleanup retained topics to MQTTAbstraction

hikinggrass:
- Track all retained topics via the MQTTAbstraction and allow all of them the be cleared at once
- Clear all topics that are published as retained during startup automatically, unless `--retain-topics` is passed to the manager (which can be useful for inspecting those topics)

### DIFF
--- a/include/utils/mqtt_abstraction.hpp
+++ b/include/utils/mqtt_abstraction.hpp
@@ -65,6 +65,10 @@ public:
     void unsubscribe(const std::string& topic);
 
     ///
+    /// \copydoc MQTTAbstractionImpl::clear_retained_topics()
+    void clear_retained_topics();
+
+    ///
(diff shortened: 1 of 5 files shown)


include/utils/mqtt_abstraction.hpp
```diff
@@ -64,6 +64,10 @@ class MQTTAbstraction {
     /// \copydoc MQTTAbstractionImpl::unsubscribe(const std::string&)
     void unsubscribe(const std::string& topic);
 
+    ///
+    /// \copydoc MQTTAbstractionImpl::clear_retained_topics()
+    void clear_retained_topics();
```
hikinggrass, line 69: TODO: maybe track the retained topics separately and move this function out of the MQTT abstraction

corneliusclaussen (approved):

PR #237, opened 2025-01-15
Revert "feature: add signal handler for termination"

a-w50:
Reverts EVerest/everest-framework#236

- using `exit()` for enforcing dumping the gcov statistics doesn't always work and has some other flaws
- instead, `__gcov_dump` should be called in the handler, but this needs to be done in everest-core, as everest-framework shouldn't know if everest-core is built for code coverage


hikinggrass (approved):

hikinggrass:
Needs a sign-off, git revert doesn't add that by default

PR #238, opened 2025-01-16
Fix optional interface handling in Rust

dorezyuk:
The framework may return null if the interface is not used (for an optional interface) - so it seems at least. This fixes the handling


andistorm:
@dorezyuk, could you change the title please to somewhat more descriptive?

I don''t really understand the PR, since I am not into this area, but at least I can't see it is changing any testing related code

corneliusclaussen:
Please update the commit description. If we merge it as is it will just say "some testing" in the git history ;-)

hikinggrass (approved):

PR #239, opened 2025-01-28
error-framework: refactor usage of clear_error

hikinggrass:


hikinggrass:
@andistorm I think your changes to this PR look good :sparkler:

everestjs/everestjs.cpp
andistorm, line 1: Js keeps its API? I agree to change as less as possible in Js :)

andistorm (approved):
LGTM 🐧

everestjs/everestjs.cpp
hikinggrass: js and rust keep their API, this only affects python and c++

PR #240, opened 2025-02-03
Fix Rust interface

dorezyuk:
* Inits the logger again before the module is ready
* makes clear ownership of the Module (now in cpp)
* fixes the new cli args (writing)
* fixes the deserialization of the interfaces coming from mqtt
* fixes the lifetimes between c++ and rust

### DIFF
--- a/everestrs/everestrs-build/src/schema/interface.rs
+++ b/everestrs/everestrs-build/src/schema/interface.rs
@@ -21,6 +21,9 @@ pub struct Interface {
 /// contrary to the definition inside the yaml file...
 #[derive(Debug, Deserialize, Serialize)]
 pub struct InterfaceFromEverest {
+    // Note: EVerest config over mqtt does not return descriptions even so
+    // they should be necessary.
+    #[serde(default)]
     pub description: String,
     #[serde(default)]
     pub cmds: BTreeMap<String, Command>,
@@ -31,6 +34,9 @@ pub struct InterfaceFromEverest {
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct Command {
+    // Note: EVerest config over mqtt does not return descriptions even so
+    // they should be necessary.    
+    #[serde(default)]
     pub description: String,
     #[serde(default)]
     pub arguments: BTreeMap<String, Type>,
--- a/everestrs/everestrs/Cargo.toml
+++ b/everestrs/everestrs/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.20.0"
 edition = "2021"
 
 [dependencies]
-argh = "0.1.10"
+clap = { version = "4.5.27", features = ["derive"] }
 cxx = { version = "1.0.107", features = ["c++17"] }
 everestrs-build = { workspace = true }
 log = { version = "0.4.20", features = ["std"] }
--- a/everestrs/everestrs/src/everestrs_sys.cpp
+++ b/everestrs/everestrs/src/everestrs_sys.cpp
@@ -10,6 +10,7 @@
 #include <cstdlib>
 #include <filesystem>
 #include <fstream>
+#include <mutex>
 #include <stdexcept>
 #include <type_traits>
 #include <variant>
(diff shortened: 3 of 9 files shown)


everestrs/everestrs/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.20.0"
 edition = "2021"
 
 [dependencies]
-argh = "0.1.10"
+clap = { version = "4.5.27", features = ["derive"] }
```
dorezyuk, line 7: argh does not support the renaming of args into camel case. So we either should use "some-thing" instead of "some_thing" or -  as done here change to clap

everestrs/everestrs/src/everestrs_sys.cpp
```diff
@@ -69,37 +70,58 @@ inline ConfigField get_config_field(const std::string& _name, int _value) {
 
 } // namespace
 
+/// @brief The flag which prevents us from re-initializing the module.
+std::once_flag mod_flag;
+
+/// @brief The central handle to the EVerest infrastructure.
+std::unique_ptr<Module> mod;
```
dorezyuk, line 77: interface before was a little bit dangerous - this fixes it and makes sure that the module is init exactly once from the user code

everestrs/everestrs/src/lib.rs
```diff
@@ -318,43 +325,38 @@ unsafe impl Send for ffi::Module {}
...
 
     /// MQTT broker port
-    #[argh(option)]
-    pub mqtt_broker_port: String, // TODO: int?
+    #[arg(long="mqtt_broker_port")]
+    pub mqtt_broker_port: u32,
 
     /// MQTT EVerest prefix
-    #[argh(option)]
+    #[arg(long="mqtt_everest_prefix")]
```
dorezyuk: renaming to match c++'s camel case

everestrs/everestrs-build/src/schema/interface.rs
```diff
@@ -31,6 +34,9 @@ pub struct InterfaceFromEverest {
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct Command {
+    // Note: EVerest config over mqtt does not return descriptions even so
+    // they should be necessary.    
+    #[serde(default)]
```
dorezyuk, line 39: the descriptions are now omitted by mqtt

everestrs/everestrs/src/lib.rs
```diff
@@ -394,40 +402,28 @@ pub trait Subscriber: Sync + Send {
     fn on_ready(&self) {}
 }
 
-/// The [Runtime] is the central piece of the bridge between c++ and Rust. We
-/// have to ensure that the `cpp_module` never outlives the [Runtime] object.
-/// This means that the [Runtime] **must** take ownership of `cpp_module`.
+/// The [Runtime] is the central piece of the bridge between c++ and Rust. The
```
dorezyuk, line 405: the old code made sure rust own the c++ module. this was necessary since we passed the refs of the runtime into the c++ module to be called in its callbacks. Now we have a static lifetime of the c++ module which means it for sure outlifes the rust code. Therefore I've changed also the lifetime ouf the rust module to static.

for the subscriber there is a similar logic. We allowed the user to drop the subscriber. This would drop the rust runtime and with the c++ module. Now, if the user drops their subscriber, the weak ref would be empty, triggering a panic. Hence, since everything is now static lifetime, we must forbid that the subscriber goes out of scope

everestrs/everestrs-build/src/schema/interface.rs
hikinggrass: Were these descriptions actually needed by the rust bindings for a feature? I know there are multiple instances where they are not yet omitted, but I would probably strip them by default in a future version everywhere

dorezyuk: no but I've written the schema according to the docs for the commands and interfaces

hikinggrass (approved):

PR #241, opened 2025-02-04
Re-introduce --config parameter for C++ Modules for compatibility

hikinggrass:
This was removed since it wasn't needed anymore since configs are distributed via MQTT It is re-introduced purely for compatibility reasons

Additionally catch the exception that can be thrown during command line parsing and exit cleanly

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14)
 
 project(everest-framework
-    VERSION 0.20.0
+    VERSION 0.20.1
     DESCRIPTION "The open operating system for e-mobility charging stations"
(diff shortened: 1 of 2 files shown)


andistorm:
For reference:

It was removed in this commit: https://github.com/EVerest/everest-framework/commit/72dec12ed244a0ee66616cee728e6ef30ea9110d

andistorm (approved):
LGTM 🐧

florinmihut (changes requested):
I would keep the config away but drop it with a warning if provided.
Moreover, we should catch the exception if a command line argument is provided that we do not support (right now we kind of crush).

florinmihut (approved):

PR #244, opened 2025-02-05
Bazel: add everest_env and everest_test

dorezyuk:
The pr adds a bazel feature to create environments and integration tests. 
An `everest_env` receives all modules and the `config.yaml` and creates a chroot. This can be then run with `bazel run` which will spawn the manager with the correct arguments.

An `everest_test` is basically the same but receives also an arbitrary binary which gets spawned after the manager. This is for now not very sophisticated but allows to run (for now simple) integration tests. 

I've used this to create a small integration test for the rust bindings to help with the development and provide immediate feedback


SirVer (approved):

everestrs/tests/types/example.yaml
```diff
@@ -0,0 +1 @@
+e
```
SirVer: @dorezyuk That seems random?

dorezyuk: yes true - will place something more descriptive here

PR #245, opened 2025-02-06
Fix config parsing for everestpy

hikinggrass:
https://github.com/EVerest/everest-framework/pull/234 changed the format of the config passed to a module which wasn't properly implemented in everestpy


everestpy/src/everest/misc.cpp
```diff
@@ -53,6 +53,14 @@ static Everest::MQTTSettings get_mqtt_settings_from_env() {
...
 }
 
+Everest::json convert_to_config_map(const Everest::json& json_config) {
+    json config_map;
+    for (auto& entry : json_config.items()) {
+        config_map[entry.key()] = entry.value().at("value");
+    }
+    return config_map;
+}
+
```
andistorm: Does anything speak against at using this existing conversion function instead?
https://github.com/EVerest/everest-framework/blob/main/everestjs/conversions.cpp#L57

SebaLukas (approved):
Looking good :+1: works on my machine

everestpy/src/everest/misc.cpp
hikinggrass: I've refactored this :)

lib/conversions.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+nlohmann::json typed_json_map_to_config_map(const nlohmann::json& typed_json_config) {
+    nlohmann::json config_map;
+    for (auto& entry : typed_json_config.items()) {
+        config_map[entry.key()] = entry.value().at("value");
+    }
+    return config_map;
+}
+} // namespace conversions
+} // namespace Everest
```
andistorm:
```suggestion
} // namespace Everest

```

andistorm (changes requested):
🐧👍

lib/conversions.cpp
hikinggrass: I've actually now moved it into the config.cpp because it's relatively specialized to some internals of the config representation (that could be refactored in an upcoming PR)

andistorm (approved):
🐧

AssemblyJohn:
Tested with sil-d20 and it works on my machine too.

PR #248, opened 2025-02-25
Fix default logging file path

florinmihut:
* with prefix is: /usr/etc/everest
* without the prefix: /etc/everest

Until now it was not an executed code path since the config file was provided as argument thus parsed instead of default values taken.

### DIFF
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -714,10 +714,14 @@ bool ModuleLoader::parse_command_line(int argc, char* argv[]) {
             assert_file(command_line_logging_config_file, "Command line provided logging config");
 
     } else {
-        auto default_logging_config_file = assert_dir(defaults::PREFIX, "Default prefix") /
-                                           fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE /
-                                           defaults::LOGGING_CONFIG_NAME;
+        auto default_logging_config_file =
+            fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE / defaults::LOGGING_CONFIG_NAME;
 
+        if (std::strcmp(defaults::PREFIX, "/usr") != 0) {
+            default_logging_config_file = defaults::PREFIX / default_logging_config_file;
+        } else {
+            default_logging_config_file = fs::path("/") / default_logging_config_file;
+        }
         this->logging_config_file = assert_file(default_logging_config_file, "Default logging config");
     }
 


lib/runtime.cpp
```diff
@@ -714,10 +714,14 @@ bool ModuleLoader::parse_command_line(int argc, char* argv[]) {
...
-                                           fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE /
-                                           defaults::LOGGING_CONFIG_NAME;
+        auto default_logging_config_file =
+            fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE / defaults::LOGGING_CONFIG_NAME;
 
+        if (std::strcmp(defaults::PREFIX, "/usr") != 0) {
+            default_logging_config_file = defaults::PREFIX / default_logging_config_file;
+        } else {
+            default_logging_config_file = fs::path("/") / default_logging_config_file;
+        }
```
andistorm, line 724: Shouldn't the prefix be added in any case?

andistorm: okay, tried to research, found out that this kind of prefix logic is used at mutliple places in this file, seems to be necessary, but I don't get why. Can you explain @hikinggrass?

Furthermore I think https://github.com/EVerest/everest-framework/blob/afab6d8e769c10de04065fe3293fff7f507c6a82/lib/runtime.cpp#L428-L429 should be moved to be done before parsing command line or at least in the beginning of parsing, since there the default_logging_config_file path is already calcualted: https://github.com/EVerest/everest-framework/blob/afab6d8e769c10de04065fe3293fff7f507c6a82/lib/runtime.cpp#L428-L429

andistorm: Which is probably out of scope of this PR

florinmihut: SYSCONF_DIR: /etc, if ${PREFIX}==/usr, otherwise ${PREFIX}/etc - is the definition in runtime.hpp.
So, no we should not add it in any case.
For the other stuff, you might be right, refactoring is necessary, but please do this in a different PR, this is just a bug fix that stops any bingup from working.

andistorm: Okay found out, that is prefix logic comes from cmake: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#special-cases

andistorm (approved):
🐧 LGTM

PR #249, opened 2025-03-06
Wait for "ready_received" before calling user code

dorezyuk:
I assume that the every module should receive `on-ready` before any other call. And once it received `on-ready` it can start calling other modules. 

Right now this logic suffers from a race condition where module `A` might receive it's `on-ready` before module `B` and start calling Rpcs/raising/clearing errors/publishing vars to module `B` - which is still waiting for its `on-ready`. The Pr addresses that for vars and errors - the cmd handling cannot be fixed without the error support Kai is working on since otherwise we might end up in a deadlock where both modules blockingly call each other's cmds from their `on-ready`. The proper way would be here to return an error


SirVer (approved):
Nice catch. Looks good for me for Rust, but I think somebody from Pionix should have a say here too.

lib/everest.cpp
```diff
@@ -486,6 +486,13 @@ void Everest::subscribe_var(const Requirement& req, const std::string& var_name,
         EVLOG_verbose << fmt::format(
             "Incoming {}->{}", this->config.printable_identifier(requirement_module_id, requirement_impl_id), var_name);
 
+        // A race condition where some other module might received its on-ready
+        // and this module not (and calling us).
+        while (!ready_received) { // In C++20 we might mark it as [[unlikely]]
+            EVLOG_warning << "Not yet ready";
+            std::this_thread::sleep_for(std::chrono::seconds(1));
```
hikinggrass: maybe we can reduce this sleep time here to something in the hundreds-of-ms range and at least store it in a constant that can be reused for the other subscriptions. making it configurable (and maybe considering a timeout) would be perfect

lib/everest.cpp
```diff
@@ -486,6 +486,13 @@ void Everest::subscribe_var(const Requirement& req, const std::string& var_name,
         EVLOG_verbose << fmt::format(
             "Incoming {}->{}", this->config.printable_identifier(requirement_module_id, requirement_impl_id), var_name);
 
+        // A race condition where some other module might received its on-ready
+        // and this module not (and calling us).
+        while (!ready_received) { // In C++20 we might mark it as [[unlikely]]
```
hikinggrass: better move this into a function that can be reused in the other cases its needed

lib/everest.cpp
```diff
@@ -486,6 +486,13 @@ void Everest::subscribe_var(const Requirement& req, const std::string& var_name,
...
+            EVLOG_warning << "Not yet ready";
```
hikinggrass: this warning could be a bit more descriptive

lib/everest.cpp
```diff
@@ -801,6 +812,8 @@ void Everest::handle_ready(const json& data) {
         on_ready_handler();
     }
 
+    this->ready_received = true;
```
hikinggrass, line 818: I'd argue "ready received" happens as soon as the message is received, this would be more of a "ready handler called". maybe we need 2 vars to track this here

hikinggrass (changes requested):

lib/everest.cpp
```diff
@@ -486,6 +486,10 @@ void Everest::subscribe_var(const Requirement& req, const std::string& var_name,
         EVLOG_verbose << fmt::format(
             "Incoming {}->{}", this->config.printable_identifier(requirement_module_id, requirement_impl_id), var_name);
 
+        // A race condition where some other module might received its on-ready
+        // and this module not (and calling us).
```
hikinggrass:
```suggestion
        // Prevent a race condition where some other module might have received its on-ready
        // already, starts publishing variables and this module is not ready yet
```

lib/everest.cpp
```diff
@@ -556,6 +560,8 @@ void Everest::subscribe_error(const Requirement& req, const error::ErrorType& er
             return;
         }
 
+        ensure_ready();
+
```
hikinggrass:
```suggestion
        // Prevent a race condition where some other module might have received its on-ready
        // already, starts publishing errors and this module is not ready yet
        ensure_ready();
```

lib/everest.cpp
```diff
@@ -823,6 +838,9 @@ void Everest::provide_cmd(const std::string& impl_id, const std::string& cmd_nam
     const auto wrapper = [this, cmd_topic, impl_id, cmd_name, handler, cmd_definition](const std::string&, json data) {
         BOOST_LOG_FUNCTION();
 
+        // TODO(ddo) Add here the ready_recieved check once we can report
```
hikinggrass:
```suggestion
        // TODO(ddo) Add here the ready_received / ensure_ready() check once we can report
```

hikinggrass (changes requested):

hikinggrass (approved):

PR #252, opened 2025-04-07
Optimize config distribution via MQTT

hikinggrass:
- Create module_config_cache directly in module Config
The information is already available to the module and does not have to be transmitted via MQTT redundantly
- Only distribute schemas via MQTT when validate_schema is set to true in settings
- Remove redundant "config_implementation" and "config_module" entries
The content of these is already provided via the "config_maps" entry in the distributed configuration for each module
- Do not distribute redundant error_types_map
Create it in modules from the interface descriptions
- Remove redundant module_provides from MQTT config distribution
This can be computed by each module from the available data

### DIFF
--- a/lib/config.cpp
+++ b/lib/config.cpp
@@ -440,11 +440,6 @@ const json& ConfigBase::get_types() const {
     return this->types;
 }
 
-std::unordered_map<std::string, ConfigCache> ConfigBase::get_module_config_cache() {
-    BOOST_LOG_FUNCTION();
-    return this->module_config_cache;
-}
-
 std::unordered_map<std::string, std::string> ConfigBase::get_module_names() {
     return this->module_names;
 }
@@ -556,7 +551,6 @@ std::optional<Mapping> ConfigBase::get_3_tier_model_mapping(const std::string& m
 void ManagerConfig::load_and_validate_manifest(const std::string& module_id, const json& module_config) {
     const std::string module_name = module_config.at("module");
 
-    this->module_config_cache[module_id] = ConfigCache();
     this->module_names[module_id] = module_name;
     EVLOG_debug << fmt::format("Found module {}, loading and verifying manifest...", printable_identifier(module_id));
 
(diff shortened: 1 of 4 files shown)


lib/config.cpp
```diff
@@ -1310,6 +1305,42 @@ json Config::get_interface_definition(const std::string& interface_name) const {
     return this->interface_definitions.value(interface_name, json());
 }
 
+void Config::populate_module_config_cache() {
+    for (const auto& [module_id, module_name] : this->module_names) {
+        this->module_config_cache[module_id] = ConfigCache();
+        const std::set<std::string> provided_impls = Config::keys(this->manifests.at(module_name).at("provides"));
+        this->interfaces[module_name] = json({});
+        this->module_config_cache[module_name].provides_impl = provided_impls;
```
Pietfried, line 1313: I think this must be 
```suggestion
        this->module_config_cache[module_id].provides_impl = provided_impls;
```

lib/config.cpp
```diff
@@ -1310,6 +1305,42 @@ json Config::get_interface_definition(const std::string& interface_name) const {
...
+        for (const auto& impl_id : provided_impls) {
+            auto intf_name =
+                this->manifests.at(module_name).at("provides").at(impl_id).at("interface").get<std::string>();
+            this->interfaces[module_name][impl_id] = intf_name;
+            this->module_config_cache[module_name].cmds[impl_id] = this->interface_definitions.at(intf_name).at("cmds");
```
Pietfried, line 1318: Same here
```suggestion
            this->module_config_cache[module_id].cmds[impl_id] = this->interface_definitions.at(intf_name).at("cmds");
```

lib/config.cpp
hikinggrass: fixed, it's actually `module_name` that's needed everywhere

lib/config.cpp
hikinggrass: fixed, it's actually `module_name` that's needed everywhere

Pietfried (approved):

PR #253, opened 2025-04-10
Change CI to reusable workflow

hikinggrass:


.ci/build-kit/scripts/compile.sh
```diff
@@ -7,11 +7,23 @@ set -e
 # source "$HOME/.cargo/env"
 
 cmake \
-    -B build \
+    -B "$EXT_MOUNT/build" \
     -S "$EXT_MOUNT/source" \
     -G Ninja \
     -DEVC_ENABLE_CCACHE=1 \
-    -DBUILD_TESTING=ON \
```
andistorm, line 14: Shouldn't this be turned on?

.gitignore
```diff
@@ -1,6 +1,6 @@
 *build
 *build-cross
-build*
+/build*
```
andistorm, line 3: Is this necessary for the CI?
I am also fine with fixing such line in this PR in case of it beeing independent.

andistorm (changes requested):

.ci/build-kit/scripts/compile.sh
hikinggrass: absolutely

.gitignore
hikinggrass: I just noticed that the new files in `.ci/build-kit` were accidentally ignored by this line. One could argue that we could also just allow `/.ci/build-kit`. But since the `/build*` entry is just there so people with in-source build-dirs don't accidentally commit things I'd say this one is enough

andistorm (approved):
🐧

PR #255, opened 2025-04-21
Migrate to Bazel 8

dorezyuk:
Goes together with 
https://github.com/EVerest/everest-utils/pull/187
https://github.com/EVerest/everest-core/pull/1180


BUILD.bazel
```diff
@@ -94,12 +94,16 @@ cc_binary(
...
         "@com_github_pboettch_json-schema-validator//:json-schema-validator",
         "@everest-framework//:controller-ipc",
         "@everest-framework//:framework",
-        "@//third-party/bazel:libcap",
+        "@libcap",
     ],
 )
+
+exports_files([
+   "dependencies.yaml",
```
hikinggrass, line 108: is this still used when the dependencies.yaml parsing is dropped for now?

MODULE.bazel
```diff
@@ -0,0 +1,119 @@
+module(
+    name = "everest-framework",
+    version = "0.0.0",
```
hikinggrass: we should add this to the [check_version_consistency.py](https://github.com/EVerest/everest-framework/blob/main/.ci/build-kit/scripts/check_version_consistency.py) to keep it in sync with what we have in CMake.
I can do that quickly for this PR

BUILD.bazel
hikinggrass: (ok, still used in everest-core)

hikinggrass (approved):

PR #256, opened 2025-04-24
Support for persistent storage and retrieval of EVerest config in SQLite

Pietfried:
This PR adds support for storing and retrieving an EVerest config in a persistant SQLite database.

* Added everest-sqlite dependency for implementation of sqlite config storage
* Added sqlite database schema in 1_up-initial.sql
* Added generic storage.hpp interface and sqlite implementation of it
* Added sqlite implementation test cases

Note: The SQLite configuration database support is not yet used in the everest-framework. This will follow in an upcoming PR.


lib/CMakeLists.txt
```diff
@@ -65,12 +66,26 @@ target_link_libraries(framework
...
         ryml::ryml
 )
+
+include(${everest-sqlite_SOURCE_DIR}/cmake/CollectMigrationFiles.cmake)
+
+message(${CMAKE_INSTALL_DATADIR}/migrations)
+
+collect_migration_files(
+     LOCATION ${PROJECT_SOURCE_DIR}/cmake/migrations
+     INSTALL_DESTINATION ${CMAKE_INSTALL_PREFIX}/migrations
```
Pietfried: rather `.../share/everest/migrations`

cmake/migrations/1_up-initial.sql
Pietfried, line 1: maybe `schemas/` is a better directory for this file? @hikinggrass

dorezyuk (approved):

dependencies.yaml
```diff
@@ -78,3 +78,8 @@ pybind11_json:
   git: https://github.com/pybind/pybind11_json.git
   git_tag: 0.2.13
   cmake_condition: "EVEREST_ENABLE_PY_SUPPORT"
+everest-sqlite:
+  git: https://github.com/EVerest/everest-sqlite.git
+  git_tag: v0.1.0
```
hikinggrass:
```suggestion
  git_tag: v0.1.1
```

lib/config/sqlite_storage.cpp
```diff
@@ -0,0 +1,694 @@
...
+    settings.run_as_user = stmt->column_text(21);
+
+    // integer
+    settings.controller_port = stmt->column_int(11);
+    settings.controller_rpc_timeout_ms = stmt->column_int(12);
+    settings.mqtt_broker_port = stmt->column_int(15);
+
+    // boolean
+    settings.telemetry_enabled = stmt->column_int(19) != 0;
+    settings.validate_schema = stmt->column_int(20) != 0;
```
hikinggrass: These column indices are pretty hard to keep in your head while reading this code, maybe we should use constants for this

Pietfried: defined constants for it

hikinggrass (approved):

PR #258, opened 2025-04-27
Set ready_received before calling on_ready(), partially reverts #249

hikinggrass:
This still ensures that handlers in modules only receive variable subscriptions and cmd calls after the ready signal has been received on framework level, as implemented in #249.

Delaying ready_received until after on_ready() has been called would lead to some modules that depend on each other to hang. While this slightly relaxes the prevention of the observed race-condition, the window is still narrower than before and we can still ensure that every module has successfully performed its init() and therefore should be ready to receive commands.

Add a new settings (and per-module override) `ensure_ready`:
This controls the behavior of the framework to either:
- ensure that the module has returned from its init and ready functions before any variable subscription callbacks are called.
- enable or disable this behavior on a per-module basis

By default this setting is turned off, reflecting the old behavior of the framework

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14)
 
 project(everest-framework
-    VERSION 0.20.3
+    VERSION 0.20.4
     DESCRIPTION "The open operating system for e-mobility charging stations"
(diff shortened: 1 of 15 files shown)


hikinggrass:
closed in favor of #259

PR #259, opened 2025-04-29
Partially revert #249

hikinggrass:
Introduce an additonal ready_processed flag

Make ensure_ready public so that language bindings that need it can make use of it

Remove ensure_ready() calls from subscribe_error/var()


dorezyuk:
@hikinggrass I've took it from here and fixed the tests in #261

Pietfried (approved):

hikinggrass:
closed in favor of #261

PR #260, opened 2025-04-29
Add script (and ci job) to check if various version numbers are in sync

hikinggrass:
everest-framework has a few sub-projects that contain their own version numbers that should be in sync with the one defined in the toplevel CMakeLists.txt

- everest-framework: `CMakeLists.txt`
- everestrs: `package.json`
- everestpy: `__init__.py `and `everestpy.cpp`
- everestrs and everestrs-build: `Cargo.toml` files

### DIFF
--- a/.ci/build-kit/scripts/check_version_consistency.py
+++ b/.ci/build-kit/scripts/check_version_consistency.py
@@ -174,8 +174,8 @@ def main() -> int:
         set_cargo_toml_version(everestrs_cargo_toml_path, new_version)
         set_cargo_toml_version(everestrs_build_cargo_toml_path, new_version)
 
-    cpp_version = get_cmake_project_version(cmakelists_txt_path)
-    print(f'C++ version: {cpp_version}')
+    everest_framework_version = get_cmake_project_version(cmakelists_txt_path)
+    print(f'everest-framework version: {everest_framework_version}')
 
     everestjs_version = get_js_package_version(package_json_path)
     print(f'everestjs version: {everestjs_version}')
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -14,6 +14,19 @@ on:
...
+    name: Ensure versions are consistent
+    runs-on: ${{ inputs.runner || 'ubuntu-22.04' }}
+    defaults:
+      run:
+        working-directory: ./source
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: source
```
andistorm, line 27: In the first moment written like this I would expect the checkout action to checkout in `./source/source` . So I would prefer setting working-directory explicitly for ` Run version consistency check` But this is just a matter of taste

.ci/build-kit/scripts/check_version_consistency.py
```diff
@@ -0,0 +1,208 @@
...
+
+    everestrs_version = get_cargo_toml_version(everestrs_cargo_toml_path)
+    print(f'everestrs version: {everestrs_version}')
+
+    everestrs_build_version = get_cargo_toml_version(everestrs_build_cargo_toml_path)
+    print(f'everestrs-build version: {everestrs_build_version}')
+
+    if all(value == cpp_version for value in [everestjs_version, everestpy_python_version, everestpy_cpp_version, everestrs_version, everestrs_build_version]):
+        return 0
+    print('Version mismatch, you can fix this by calling this script with --fix which will apply the version defined in the toplevel CMakeLists.txt')
```
andistorm: Above the version is named/listed as `C++ version`. Intuitive I wouldn't map the term `C++ version` to the term `version defined in toplevel CMakeLists.txt`, but it would be nice to directly understand which value is used with `--fix`

andistorm: Maybe the hint should suggest using `--update-version` with the intended version instead

andistorm (changes requested):

.ci/build-kit/scripts/check_version_consistency.py
hikinggrass: I've renamed it to _everest-framewok version_ and added an additional hint

.github/workflows/build_and_test.yaml
hikinggrass: I just oriented myself on the workflow(s) in everest-ci, but I'm open for alternatives :smile:

andistorm (approved):
🐧

PR #262, opened 2025-05-02
Rust: Add ignoring of variables and errors

dorezyuk:
Currently Rust eagerly subscribes to all vars and error topics. This can increase the number of threads used by the system. With this feature the user can opt out from subscribing to certain topics

### DIFF
--- a/everestrs/everestrs/src/everestrs_sys.cpp
+++ b/everestrs/everestrs/src/everestrs_sys.cpp
@@ -154,8 +154,7 @@ void Module::subscribe_all_errors(const Runtime& rt) const {
     const auto manifest = config_->get_manifests().at(module_name);
     const auto requires = manifest.at("requires");
     for (const Requirement& req : config_->get_requirements(module_id_)) {
-        if (requires.at(req.id).contains("ignore") &&
-            requires.at(req.id).at("ignore").contains("errors") &&
+        if (requires.at(req.id).contains("ignore") && requires.at(req.id).at("ignore").contains("errors") &&
             requires.at(req.id).at("ignore").at("errors").get<bool>()) {
             continue;
         }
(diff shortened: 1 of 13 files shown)


SirVer:
@dorezyuk I will not be able to review this in the next 7-10 business days. If you need it earlier, please ask a team mate.

wku12:
I'm fine with those changes

schemas/manifest.yaml
```diff
@@ -109,6 +109,19 @@ properties:
             type: integer
             minimum: 1
             default: 1
+          ignore:
```
dorezyuk, line 113: add comment here this is only for rust

lib/everest.cpp
```diff
@@ -593,7 +601,7 @@ std::shared_ptr<error::ErrorFactory> Everest::get_error_factory(const std::strin
 
 std::shared_ptr<error::ErrorManagerReq> Everest::get_error_manager_req(const Requirement& req) {
     if (this->req_error_managers.find(req) == this->req_error_managers.end()) {
-        EVLOG_error << fmt::format("Error manager for {} not found!", req.id);
+        EVLOG_info << fmt::format("Error manager for {} not found", req.id);
```
dorezyuk: throw the exception instead of error print

everestrs/everestrs/src/everestrs_sys.cpp
```diff
@@ -151,8 +151,16 @@ void Module::subscribe_variable(const Runtime& rt, rust::String implementation_i
 
 void Module::subscribe_all_errors(const Runtime& rt) const {
     for (const Requirement& req : config_->get_requirements(module_id_)) {
+        std::shared_ptr<Everest::error::ErrorManagerReq> error_manager_ptr = nullptr;
+        try {
+            error_manager_ptr = handle_->get_error_manager_req(req);
+        } catch (const std::runtime_error& ex) {
+            // This is expected if the manifest has `irgnore.errors=true`
```
hikinggrass:
```suggestion
            // This is expected if the manifest has `ignore.errors=true`
```

hikinggrass (approved):
small nit

PR #266, opened 2025-05-08
Changed parse_everest_config to parse_module_configs

Pietfried:
Changed parse_everest_config to parse_module_configs . Parsing the complete config including settings and modules is never required.

### DIFF
--- a/include/utils/config/types.hpp
+++ b/include/utils/config/types.hpp
@@ -102,7 +102,7 @@ struct ConfigurationParameter;
 struct ModuleConfig;
 using ModuleId = std::string;
 using RequirementId = std::string;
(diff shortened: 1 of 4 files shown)


hikinggrass (approved):

include/utils/config/types.hpp
```diff
@@ -102,7 +102,7 @@ struct ConfigurationParameter;
 struct ModuleConfig;
 using ModuleId = std::string;
 using RequirementId = std::string;
-using ConfigEntry = std::variant<std::string, bool, int, double>;
+using ConfigEntry = std::variant<std::string, bool, int, double, fs::path>;
```
Pietfried:
```suggestion
using ConfigEntry = std::variant<std::string, bool, int, double>;
```

PR #267, opened 2025-05-08
feat: Integrate booting EVerest from database

Pietfried:


include/framework/runtime.hpp
```diff
@@ -126,7 +133,9 @@ struct ManagerSettings {
     MQTTSettings mqtt_settings;       ///< MQTT connection settings
     RuntimeSettings runtime_settings; ///< Runtime settings needed to successfully run modules
 
-    ManagerSettings(const std::string& prefix, const std::string& config);
+    ConfigSource config_source; ///< Source of the config (file or database)
+
+    ManagerSettings(const std::string& prefix, const std::string& config, const bool boot_from_db = false);
```
Pietfried: Make `from_db` an enum

lib/runtime.cpp
```diff
@@ -51,9 +52,11 @@ ManagerSettings::ManagerSettings(const std::string& prefix_, const std::string&
     }
 
     fs::path prefix;
-    if (prefix_.length() != 0) {
+    if (!prefix_.empty()) {
         // user provided
         prefix = assert_dir(prefix_, "User provided prefix");
+    } else {
+        prefix = assert_dir(defaults::PREFIX, "Default prefix");
     }
```
Pietfried: deprecate prefix param in database and config file and print warning in case selected prefix and config or db prefix does not match

hikinggrass (approved):

PR #270, opened 2025-05-16
Set rust-version to 1.74 to ensure Cargo.lock version 3 is used

hikinggrass:
1.74. is the Rust version which is in the WORKSPACE.bazel, this doesn't yet understand Cargo.lock version 4 A more recent system cargo would set version = 4 when running cargo generate-lockfile unless a appropriate rust-version is set


SirVer:
Defering to @dorezyuk

hikinggrass:
closed in favor of #255

PR #271, opened 2025-05-26
Fix some compiler warnings

mhei:
This fixes some compiler warnings which were straight-forward to fix.

### DIFF
--- a/src/controller/controller.cpp
+++ b/src/controller/controller.cpp
@@ -16,7 +16,7 @@
 
 using json = nlohmann::json;
 
-int main(int argc, char* argv[]) {
+int main([[maybe_unused]] int argc, char* argv[]) {
 
     if (strcmp(argv[0], MAGIC_CONTROLLER_ARG0)) {
         fmt::print(stderr, "This binary does not yet support to be started manually\n");
--- a/lib/error/error_database_map.cpp
+++ b/lib/error/error_database_map.cpp
@@ -40,7 +40,7 @@ std::list<ErrorPtr> ErrorDatabaseMap::get_errors_no_mutex(const std::list<ErrorF
         std::function<bool(const ErrorPtr&)> pred;
         switch (filter.get_filter_type()) {
         case FilterType::State: {
-            pred = [](const ErrorPtr& error) { return false; };
+            pred = []([[maybe_unused]] const ErrorPtr& error) { return false; };
             EVLOG_error << "ErrorDatabaseMap does not support StateFilter. Ignoring.";
         } break;
         case FilterType::Origin: {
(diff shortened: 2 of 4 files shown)


src/controller/controller.cpp
```diff
@@ -16,7 +16,7 @@
 
 using json = nlohmann::json;
 
-int main(int argc, char* argv[]) {
+int main([[maybe_unused]] int argc, char* argv[]) {
```
hikinggrass, line 19: We should check if argc is < 1 here and exit then as well, like:

```cpp
int main(int argc, char* argv[]) {

    if (argc < 1 or strcmp(argv[0], MAGIC_CONTROLLER_ARG0)) {
```

hikinggrass (changes requested):

src/controller/controller.cpp
mhei: I can change it if you insist, but per standard, you always have argc >= 1 - by convention the argv[0] is usually the name/path to the program itself - or here in our case the magic value. I actually though, that it would not be possible to call `execl` with the second parameter already set to NULL. But yes, this triggers a compiler warning, but it executes:

**test.c**
```c
#include <stdio.h>
#include <unistd.h>  // for execl
#include <stdlib.h>  // for perror and exit

int main() {
    printf("Calling execl to run '/tmp/test2'\n");

    // execl(path, arg0, arg1, ..., NULL);
    // Note: arg0 is conventionally the name of the program itself
    if (execl("/tmp/test2", (char *)NULL) == -1) {
        perror("execl failed");
        exit(EXIT_FAILURE);
    }

    // This line will never be reached if execl is successful
    printf("This will not be printed if execl succeeds\n");
    return 0;
}

```

**test2.c**
```c
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

int main(int argc, char *argv[]) {
    printf("argc = %d, argv[0] = %s (strlen: %ld)\n", argc, argv[0], strlen(argv[0]));
    return 0;
}
```

**Output:**
```
mhei@kerker:/tmp$ ./test2
argc = 1, argv[0] = ./test2 (strlen: 7)

mhei@kerker:/tmp$ ./test
Calling execl to run '/tmp/test2'
argc = 1, argv[0] =  (strlen: 0)
```

I didn't expected this :smile:

hikinggrass: I was actually thinking of the case where `argc == 0` and therefore `argv[0] == NULL`, which should lead to undefined behavior in  the `strcmp()` call. However I think I missed that this [has actually been addressed a few years ago](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=dcd46d897adb70d63e025f175a00a89797d31a43). So I think your version is totally fine :smile:

hikinggrass (approved):

src/controller/controller.cpp
mhei: Nice link - thanks!

hikinggrass:
@mhei just needs a rebase on your end then we can merge this :+1:

PR #273, opened 2025-05-27
Fix multiple compiler warnings

hikinggrass:
- reordering of various members
- unused variable "requirement"
- unnecessary moves
- rename variable "requires" to "module_requires" because "requires is a C++20 keyword


Pietfried (approved):
lgtm

PR #287, opened 2025-07-22
C++: Add support for multiple "provides"

hikinggrass:
Add support for multiple "provides"

A (C++) module can opt-into this behavior, when setting the new
"quantity" option in the manifest to a value > 1.

In this case ev-cli generates handle_* and publish_* functions that take an index as first parameter


schemas/manifest.yaml
```diff
@@ -85,6 +85,10 @@ properties:
               Config set for this implementation (and possibly default
               values) declared as json schema
             $ref: '#/$defs/config_set_schema'
+          quantity:
+            type: integer
+            minimum: 0
```
dorezyuk, line 90: minimum should be 1?

include/framework/everest.hpp
```diff
@@ -74,6 +79,7 @@ class Everest {
     /// \brief Allows a module to indicate that it provides the given command \p cmd
     ///
     void provide_cmd(const std::string& impl_id, const std::string& cmd_name, const JsonCommand& handler);
+    void provide_cmd(const std::string& impl_id, const std::string& cmd_name, const JsonCommandMultiple& handler);
```
dorezyuk, line 82: should we just have this as the generic signature and just change the generated `handle_{cmd}` signature based on the quantity parameter?

include/framework/everest.hpp
```diff
@@ -87,26 +93,32 @@ class Everest {
     ///
     void publish_var(const std::string& impl_id, const std::string& var_name, nlohmann::json value);
 
+    ///
+    /// \brief Publishes a variable at the given \p _index, of the given \p impl_id, names \p var_name with the given \p
+    /// value
+    ///
+    void publish_var(std::size_t index, const std::string& impl_id, const std::string& var_name, nlohmann::json value);
```
dorezyuk, line 100: same

include/framework/everest.hpp
```diff
@@ -87,26 +93,32 @@ class Everest {
...
     /// var_name. The given \p callback is called when a new value becomes available
     ///
     void subscribe_var(const Requirement& req, const std::string& var_name, const JsonCallback& callback);
 
     ///
-    /// \brief Return the error manager for the given \p impl_id
+    /// \brief Return the error managers for the given \p impl_id
     ///
-    std::shared_ptr<error::ErrorManagerImpl> get_error_manager_impl(const std::string& impl_id);
+    std::vector<std::shared_ptr<error::ErrorManagerImpl>> get_error_manager_impl(const std::string& impl_id);
```
dorezyuk, line 111: then the signatures should be changed to `get_error_manager_impls` (also for the 3 below)

include/framework/everest.hpp
```diff
@@ -208,10 +220,11 @@ class Everest {
     std::shared_ptr<MQTTAbstraction> mqtt_abstraction;
     Config config;
     std::string module_id;
-    std::map<std::string, std::shared_ptr<error::ErrorManagerImpl>> impl_error_managers; // one per implementation
-    std::map<std::string, std::shared_ptr<error::ErrorStateMonitor>>
+    std::map<std::string, std::vector<std::shared_ptr<error::ErrorManagerImpl>>>
```
dorezyuk, line 223: should we also make it unodered?

PR #292, opened 2025-08-05
Refactor MessageHandler for Improved Thread Handling and MQTT communication

Pietfried:
Redesign of the framework message handling to reduce the number of threads per module and avoid wake-up of module threads in case messages can actually be discarded.

**Key Changes**:
* For the MQTT communication, modules now use only two threads for message processing:
   * operational_message thread (vars, commands, errors)
   * result_message thread (command results, config responses)
* Modules now execute var, command and error handlers sequentially
* Message Handlers execute without holding locks, enabling safe cross-handler registration
* MQTT Topic structure adjustmens:
  * More specific topic paths (e.g., {prefix}/cmd/{cmd_name} instead of generic {prefix}/cmd or {prefix}/var). This allows modules to only wake up and process messages they are actually interested in
  * Response routing: Direct command response routing to calling modules via dedicated response topics
* Message type enforcement: All MQTT messages now include explicit msg_type field for proper routing


Pietfried:
The changes in this PR result in the fact that it is not guaranteed anymore that different variables are delivered in the order they are published. The same order of same variables published is maintained.

Before this change it was guaranteed that variables were delivered in the same order they were published because all variables of a module were published on the same topic. The topic structure of vars (and commands) has been changed in this PR.

It must be investigated whether existing modules rely on this guarantee before this PR is merged.

hikinggrass (approved):

PR #295, opened 2025-08-21
Rust: Fix clippy warning

dorezyuk:
The derived enum names may be invalid/bad for clippy


hikinggrass (approved):

hikinggrass (changes requested):
still seems to lead to build failures

hikinggrass (approved):

SirVer (approved):

PR #297, opened 2025-09-02
Update pybind11 to 3.0.1 and fix options

hikinggrass:
This fixes cross compiling with everestpy enabled

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,7 @@ endif()
 
 # everest python wrapper
 if (EVEREST_ENABLE_PY_SUPPORT)
+    set(PYTHON_MODULE_EXTENSION ".so")
     add_subdirectory(everestpy)
 endif()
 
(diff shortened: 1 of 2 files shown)


Pietfried (approved):

barsnick:
We noticed this yesterday as well, when cross-compiling 2025.7.0 for the first time. (And incidentally fixed it in our CMake toolchain file in the same way.) Thanks for this fix. 👍🏻

When and how did this break? Does it need to be back-ported to a 2025.7.1?

hikinggrass:
> When and how did this break? 

The cross-compile issue seems to have appeared with the update from pybind11 2.11.1 (which was working) to 2.13.6 in #290 so should only affect 2025.7.0 . The python lib accidentally containing the host architecture had been worked around in meta-everest for quite some time, so this might have been broken for longer.

> Does it need to be back-ported to a 2025.7.1?

If we do a 2025.7.1 then there are also a few other things in everest-core and libocpp that could be backported, lets see after 2025.8.0 is out what could be included

PR #299, opened 2025-09-25
Update libwebsockets dependency to 4.4.1

barsnick:
This updates the libwebsockets library to 4.4.1.

This also adapts the controller server to an incompatible API change in struct `lws_http_mount`. This change adapts to both old and new API.

NOTE: Keep the dependency aligned with everest-core (incoming via RpcApi PR https://github.com/EVerest/everest-core/pull/1324).


hikinggrass:
Looks good from my perspective, we also have to update this in [libocpp](https://github.com/EVerest/libocpp) as well then

hikinggrass (approved):

PR #303, opened 2025-10-17
Update fmt to 12.1.0 and set FMT_SYSTEM_HEADERS cmake var to ON

hikinggrass:
There was already an issue with fmt <12 preventing out-of-the box builds with newer version of clang


andistorm (approved):
LGTM 🐧

PR #310, opened 2025-11-05
C++: Suppress recurrent log line

dorezyuk:
Otherwise it on every clear_all_errors call we would see 

```"Cleared 0 errors:\n```

### CODE REVIEW

File: lib/error/error_manager_impl.cpp [hunk 1]
```diff
@@ -95,6 +95,10 @@ std::list<ErrorPtr> ErrorManagerImpl::clear_error(const ErrorType& type, const E
 std::list<ErrorPtr> ErrorManagerImpl::clear_all_errors() {
     const std::list<ErrorFilter> filters = {};
     std::list<ErrorPtr> res = database->remove_errors(filters);
+    if (res.empty()) {
+        EVLOG_debug << "Errors can't be cleared, because on error is active";
```
andistorm:
```suggestion
        EVLOG_debug << "No errors can be cleared, because no errors are active";
```

andistorm (approved):
🐧 LGTM

PR #312, opened 2025-11-26
fix: Add extra check for etc path

chrizap:
Problem
When using a custom prefix ending with "usr" (e.g., /extra/usr), the etc directory path was incorrectly resolved to /extra/usr/etc instead of /extra/etc.

An extra check has been added which uses parent path as root directory for etc.

Test Cases
  Prefix:   /usr
  Got:      /etc

  Prefix:   /extra/usr
  Got:      /extra/etc

  Prefix:   /opt/usr
  Got:      /opt/etc

  Prefix:   /usr/local
  Got:      /usr/local/etc

  Prefix:   /extra
  Got:      /extra/etc

  Prefix:   /opt/myapp
  Got:      /opt/myapp/etc

### DIFF
--- a/lib/runtime.cpp
+++ b/lib/runtime.cpp
@@ -116,11 +116,13 @@ void ManagerSettings::init_settings(const everest::config::Settings& settings) {
     fs::path etc_dir;
     {
         // etc directory
-        const auto default_etc_dir = fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE;
-        if (prefix.string() != "/usr") {
-            etc_dir = prefix / default_etc_dir;
-        } else {
+        const auto default_etc_dir = (fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE).relative_path();
+        if (prefix.string() == "/usr") {
             etc_dir = fs::path("/") / default_etc_dir;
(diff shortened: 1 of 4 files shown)


hikinggrass:
Hi, I think you're missing a sign-off in your commit which means the DCO check fails, can you fix that? Apparently there's also a line that needs reformatting since clang-format complains :D thanks a lot!

lib/runtime.cpp
```diff
@@ -116,11 +116,13 @@ void ManagerSettings::init_settings(const everest::config::Settings& settings) {
...
     {
         // etc directory
-        const auto default_etc_dir = fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE;
-        if (prefix.string() != "/usr") {
-            etc_dir = prefix / default_etc_dir;
-        } else {
+        const auto default_etc_dir = (fs::path(defaults::SYSCONF_DIR) / defaults::NAMESPACE).relative_path();
+        if (prefix.string() == "/usr") {
             etc_dir = fs::path("/") / default_etc_dir;
+        } else if (etc_root.filename() == "usr") {
```
hikinggrass: Looks like this line doesn't compile, what's etc_root supposed to point to?

hikinggrass (approved):

PR #313, opened 2025-11-27
Rust: pin cxxbridge version, update to 1.0.189

IvarWithoutBones:
When building an EVerest workspace I noticed all Rust modules failed to link because none of the CXX FFI symbols could be resolved.

The issue appears to be that CMake installs an unbounded (i.e the latest) version of the `cxxbridge-cmd` tool at configure time, even though it must match the version of the `cxx` crate used in `everestrs` (https://github.com/dtolnay/cxx/issues/1507).
In release 1.0.189 (the latest as of writing) they started incorporating the version number in symbol names to detect exactly this scenario, which broke my build. CMake installed `cxxbridge-cmd` 1.0.189, but the `cxx` library version was locked to an older release.

This PR fixes that by: 
* Updating the `cxx` crate to the latest version. This doesn't solve the root cause of the issue, but its nice to have done anyways.
 * Locking the version of `cxxbridge-cmd` when installing to match `cxx`, and try to update it if the existing binary is outdated.(Re)running CMake's configure phase when updating `cxx` now fixes the issue.

It also does a few other Rust related things:
* Use the `cxxbridge-cmd` version from `$PATH` if it has the correct version. Useful for offline builds.
* Ensure we don't inherit cross compilation flags from the environment when installing `cxxbridge-cmd`.
* Document how to get `cargo build` to work in external repositories.

### DIFF
--- a/everestrs/CMakeLists.txt
+++ b/everestrs/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(CXXBRIDGE_VERSION 1.0.189) # Must be kept in sync with `everestrs/Cargo.toml`
 set(CXXBRIDGE_INSTALL_PATH ${CMAKE_CURRENT_BINARY_DIR}/cargo/bin/cxxbridge)
 
-# Checks if the cxxbridge binary at CXXBRIDGE_PATH exists matches the required version.
+# Checks if the cxxbridge binary at CXXBRIDGE_PATH exists and matches the required version.
 # If not, sets VERSION_MATCHES to FALSE.
 function(everestrs_check_cxxbridge_version CXXBRIDGE_PATH VERSION_MATCHES)
     if(NOT CXXBRIDGE_BINARY OR NOT EXISTS ${CXXBRIDGE_BINARY})
--- a/everestrs/CMakeLists.txt
+++ b/everestrs/CMakeLists.txt
@@ -1,9 +1,48 @@
-set(CXXBRIDGE_BINARY ${CMAKE_CURRENT_BINARY_DIR}/cargo/bin/cxxbridge)
-if (NOT EXISTS ${CXXBRIDGE_BINARY})
+set(CXXBRIDGE_VERSION 1.0.189) # Must be kept in sync with `everestrs/Cargo.toml`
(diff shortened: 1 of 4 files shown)


everestrs/CMakeLists.txt
```diff
@@ -1,9 +1,48 @@
-set(CXXBRIDGE_BINARY ${CMAKE_CURRENT_BINARY_DIR}/cargo/bin/cxxbridge)
-if (NOT EXISTS ${CXXBRIDGE_BINARY})
+set(CXXBRIDGE_VERSION 1.0.189) # Must be kept in sync with `everestrs/Cargo.toml`
+set(CXXBRIDGE_INSTALL_PATH ${CMAKE_CURRENT_BINARY_DIR}/cargo/bin/cxxbridge)
+
+# Checks if the cxxbridge binary at CXXBRIDGE_PATH exists matches the required version.
```
barsnick: A verb too much, or missing "and".
```suggestion
# Checks if the cxxbridge binary at CXXBRIDGE_PATH exists and matches the required version.
```

IvarWithoutBones: Thanks for the review, fixed

everestrs/CMakeLists.txt
```diff
@@ -1,9 +1,48 @@
-set(CXXBRIDGE_BINARY ${CMAKE_CURRENT_BINARY_DIR}/cargo/bin/cxxbridge)
-if (NOT EXISTS ${CXXBRIDGE_BINARY})
+set(CXXBRIDGE_VERSION 1.0.189) # Must be kept in sync with `everestrs/Cargo.toml`
+set(CXXBRIDGE_INSTALL_PATH ${CMAKE_CURRENT_BINARY_DIR}/cargo/bin/cxxbridge)
+
+# Checks if the cxxbridge binary at CXXBRIDGE_PATH exists and matches the required version.
+# If not, sets VERSION_MATCHES to FALSE.
```
dorezyuk, line 5: where do we use that?

dorezyuk: ah got it

dorezyuk (approved):

hikinggrass (approved):

PR #314, opened 2025-11-27
python: enter virtual environments before starting modules

IvarWithoutBones:
We noticed that Python dependencies are currently not isolated between modules, specifically with `pydantic`: `PyEvJosev` requires a rather old version which conflicts with our own modules. While we could upgrade Josev to temporarily fix the problem, this doesn't scale very well in the long run.

This PR makes a step towards improving the situation, it makes the manager process activate each module's virtual environment before starting it (if present). This PR does not automatically generate or install the module's dependencies in the virtual environment, I tested it by generating one by hand:
```sh
cd /libexec/everest/modules/PyEvJosev
python3 -m venv .venv
source .venv/bin/activate
python3 -m pip install -r requirements.txt
```

Ideally CMake would generate everything for us at build time, but this turns out to be a bit trickier than expected:

* Virtual environments are not portable, they contain absolute paths to themselves in various generated scripts.
* While `everest-cmake` is already able to generate a wheel for each module (which we could reasonably quickly install at runtime), it isn't able to cross-compile them. This matters because a lot of Python libraries link to native code that is fetched/compiled at install time, so forcibly using wheels generated by CMake would break EVerest's cross-compilation support.

  There sadly doesn't appear to be a standardized way to indicate a cross-compilation context when building Python packages, though we might be able to utilize projects like [crossenv](https://pypi.org/project/crossenv) or [cibuildwheel](https://github.com/pypa/cibuildwheel). This needs more investigation.
* If cross-compilation doesn't turn out to be viable we could instead install/download dependencies at runtime, from the manager process. This completely resolves any cross-compilation concerns, but in turn has the downside of slowing down module startup.

In any case, after creating a virtual environment we would need to enter it from the manager process, which this PR does (in addition to some small cleanup). I'm not sure what the best way to proceed forward is, I'd love some feedback. What would be the appropriate repository to open an issue about this?


andistorm (approved):
@IvarWithoutBones thanks for your contribution. 🐧 Looks good to me, just two things:
* I added a fix that makes use of `execvp()` instead of `execv()` this is necessary for the case js modules are started and python modules without dedicated venv are started.
* I reverted the clang-tidy fixes commit, since it is not related. We should open up a separate PR for this.

hikinggrass (approved):

andistorm:
I opened https://github.com/EVerest/everest-framework/pull/315 for the clang tidy spellchecker commit

IvarWithoutBones:
> I added a fix that makes use of execvp() instead of execv() this is necessary for the case js modules are started and python modules without dedicated venv are started.

Oh right, that makes sense. Thanks for fixing it.

> I reverted the clang-tidy fixes commit, since it is not related. We should open up a separate PR for this.

I'll keep future PRs more to the point, I appreciate you separating it out this time :)

IvarWithoutBones:
> I added a fix that makes use of execvp() instead of execv() this is necessary for the case js modules are started and python modules without dedicated venv are started.

Oh right, that makes sense. Thanks for fixing it.

> I reverted the clang-tidy fixes commit, since it is not related. We should open up a separate PR for this.

I'll keep future PRs more to the point, I appreciate you separating it out this time :)

